PR #95, opened 2021-01-13
Repeat (looping)

darth-cheney:
## What
This PR implements the basic looping constructs, called `repeat`. It is based almost entirely upon the HT specification.
  
### Repeat Blocks ###
Looping is conducted inside of what we call "repeat blocks." Each of these blocks begins with the keyword `repeat` (along with more specific forms), some lines of statements to execute each run of the loop, and the `end repeat` to finish the block. There are several types of repeat blocks one can use:
#### `repeat for` ####
This construct repeats a loop for some number of times.
```simpletalk
repeat for 5 times
    beep
end repeat
```
#### `repeat until` ####
This construct will repeat the loop until some condition evaluates to true:
```simpletalk
put 5 into myCounter
repeat until myCounter <= 0
    beep
    put (myCounter - 2) into myCounter
end repeat
```
#### `repeat while` ####
This construct is the opposite of the `until` variant, and will repeat the statements in the block as long as the conditional evluates to true
```simpletalk
put 0 into myVariable
repeat while myVariable < 5
    beep
    put (myVariable + 3) into myVariable
end repeat
```
#### `repeat with` ####
This construct specifies a range and puts the current index of that range into a named variable in the repeat line:
```simpletalk
repeat with myCounter = 4 to 12
    beep
    answer myCounter
end repeat
```
### Adjustment Statements ###
There are two kinds of "adjustment statements" one can make inside of a repeat block: `exit repeat`, which breaks the loop entirely, and `next repeat` which stops executing the current statement loop and skips to the next run of the loop.
  
## Tests ##
Included are [grammar tests](https://github.com/UnitedLexCorp/SimpleTalk/blob/eric-repeat/js/ohm/tests/test-repeat-grammar.js) and [integration/interpreter tests](https://github.com/UnitedLexCorp/SimpleTalk/blob/eric-repeat/js/objects/tests/test-repeat-with-preload.js). All are passing.

### DIFF
--- a/js/objects/tests/test-if-then-with-preload.js
+++ b/js/objects/tests/test-if-then-with-preload.js
@@ -95,7 +95,7 @@ describe('Inline if-then tests', () => {
     it('Can deal with an equality comparison (true)', () => {
         let script = [
             'on click',
-            '\tif (2+2) = 4 then set "name" to "evaluated3"',
+            '\tif (2+2) == 4 then set "name" to "evaluated3"',
             'end click'
         ].join('\n');
         compileButtonScript(script);
@@ -111,7 +111,7 @@ describe('Inline if-then tests', () => {
     it('Can deal with an equality comparison (false)', () => {
         let script = [
             'on click',
-            '\tif (2+3) = 4 then set "name" to "evaluated4"',
+            '\tif (2+3) == 4 then set "name" to "evaluated4"',
             'end click'
         ].join('\n');
         compileButtonScript(script);
@@ -300,7 +300,7 @@ describe('Singleline If-Then tests', () => {
     it('Can deal with an equality comparison (true)', () => {
         let script = [
             'on click',
-            '\tif (2+2) = 4',
+            '\tif (2+2) == 4',
             'then set "name" to "evaluated3"',
             'end click'
         ].join('\n');
@@ -317,7 +317,7 @@ describe('Singleline If-Then tests', () => {
     it('Can deal with an equality comparison (false)', () => {
         let script = [
             'on click',
-            '\tif (2+3) = 4',
+            '\tif (2+3) == 4',
             'then set "name" to "evaluated4"',
             'end click'
         ].join('\n');
--- a/js/objects/tests/test-repeat-with-preload.js
+++ b/js/objects/tests/test-repeat-with-preload.js
@@ -153,7 +153,7 @@ describe("Repeat Looping Tests", () => {
     });
 
     describe("Misc example tests", () => {
-        it("Can exit repeat on if condition", () => {
+        it("Can exit repeat on if condition (until)", () => {
             let script = [
                 "on click",
                 "put 0 into Counter",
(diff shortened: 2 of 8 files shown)


dkrasner:
This looks goods. 

A few issues:

* did you mean to add a `beep` command, or are you just using it above in examples to show what it could be, i.e. there is no `beep` command either in the grammar or anywhere else
* I made the typo below but it exposed a variable error: `"Count"` is not defined, it's actually `"Counter"` but no error is throw, nor does look forever as it should there. 
```
on click
    put 0 into Counter
    repeat while Counter < 5
       answer Counter
       put (Count + 1) into Counter
    end repeat
end click
```
(Changing `"Count"` to `"Counter"` does what you want). 
* We should change the equivalence from "=" to "=="   Overloading the assignment operator "=" is confusing and bad (math) notation, i.e. this should work
```
on click
    put 0 into Counter
    repeat until Counter == 5
       answer Counter
       put (Counter + 1) into Counter
    end repeat
end click
```
* I couldn't get this to work:
```
on click
    put 0 into Counter
    repeat until Counter = 5
       answer Counter
       put (Counter + 1) into Counter
       if Counter >= 3 then exit repeat
    end repeat
end click
```
tried various syntax option, single/multi line etc 

maybe best to fix these and then i'll play around some more

dkrasner:
PS we really do need better grammar message handling, i.e. we should underline where in the grammar it fails, why and not have it be a system wide error...

darth-cheney:
@dkrasner Have a look at the latest commits. I decided against creating domain objects within the Interpreter for the moment. The few strategies I tried ended up making things more complex in the end. I think we will have a better idea of how to optimize/redesign this once we have been cruising with a v1 for a bit.

dkrasner:
@darth-cheney the exit repeat works, which is a big. The rest of the issues ^^ are still present. I didn't go testing further, since it makes sense to resolve them before moving fwd.

darth-cheney:
I'm going to work on the `== / =` issue on this PR. As for this:
> I made the typo below but it exposed a variable error: "Count" is not defined, it's actually "Counter" but no error is throw, nor does it loop forever as it should there.
  
The issue is that `undefined` has a value in Javascript, and in cases where we look up a variable name in the current context dictionary that doesn't exist, we get an undefined rather than an error. This is where we need to make a design decision. To me the decision is obvious: we should have lookups like this throw an error. I will proceed with that in mind and if we don't like the change we can roll back whatever commit results.

dkrasner:
> I'm going to work on the `== / =` issue on this PR. As for this:
> 
> > I made the typo below but it exposed a variable error: "Count" is not defined, it's actually "Counter" but no error is throw, nor does it loop forever as it should there.
> 
> The issue is that `undefined` has a value in Javascript, and in cases where we look up a variable name in the current context dictionary that doesn't exist, we get an undefined rather than an error. This is where we need to make a design decision. To me the decision is obvious: we should have lookups like this throw an error. I will proceed with that in mind and if we don't like the change we can roll back whatever commit results.

yeah def. Should be something like "ReferenceError" then with more details, such as `type: variable` so then we could have an error handler provide local/global vars

similarly `type: handler` would allow the error handler to show available handlers in the hierarchy

darth-cheney:
@dkrasner Ok, the latest commits should deal with the open issues on this PR. Let me know if you see anything else that isn't working right

dkrasner:
@darth-cheney this all looks good

I played around for a while, couldn't break it . 

Unless you want to add a `beep` command - that's the one issue above that's remaining, I would merge it. Not sure i care much about beeping...

Issue #97, opened 2021-01-26
Adding and removing click event does not play well with Halo


dkrasner:
### Main issue ###

If a a `click` event is removed, then added either via an editor or a `respond to "click"` command, the view will respond to the click but no longer interface with the Halo. For example, `shift-rightClick` will not open or close the Halo. 

We should treat the `click` event as special. When removing it we should not truly remove it from the view but only from the underlying model.partProperties effectively preventing any sort of `on click` handling, but still allowing for the halo. And when adding it back in we should make sure to handle the `shift` button properly. 

Another option is to try to experiment with having two separate handlers, one added to the DOM element via the `element.addEventListener()` API and the other using the `element.onClick` which is currently how events are added and removed. But this requires some experimentation.

dkrasner:
no longer relevant

PR #98, opened 2021-01-26
Button editor

dkrasner:
### Main points ###

After some pain, a button editor. It does the basic things: 

Changes the name
Adjusts background and font colors (via our colorWheel)
Lists, adds, removes events
And lets you script

image

The editor is opened by pressing the "write/edit" icon, which previously opened the script/field editor. 

Instead of embedding all the widgets necessary to do editing (color, script) inside one editor view I decided that each would open it's own, absolutely positioned, widget. After some experimenting with both options, this was easier to style and scale, and in the end I prefer it overall. This way you can lay out the various sub-editor widgets as you see fit and we can just keep making more, adding  a button to open it to the main one:

image

I tried to [stick to SimpleTalk](https://github.com/UnitedLexCorp/SimpleTalk/compare/daniel-button-editor?expand=1#diff-d460e66340863153b40b711a201ef3dea6540293e5d5bc796755e4f238f0a473R241) in evoking the various sub-editors so that it would be clearer how one could build such a thing with just SimpleTalk buttons. You almost can do it at the moment... 

`editorOpen` is a now a [property](https://github.com/UnitedLexCorp/SimpleTalk/compare/daniel-button-editor?expand=1#diff-d460e66340863153b40b711a201ef3dea6540293e5d5bc796755e4f238f0a473R241) on all parts. 

#### Tests ####

I wrote as many [tests](https://github.com/UnitedLexCorp/SimpleTalk/compare/daniel-button-editor?expand=1#diff-0a1b715b2d04f9c839158dcb7132bfb2ba73b511ea8d03a2097b1851ac0dbe49) as I could think of, without dealing with npm canvas context issues (which you need to solve if you want to properly test our color wheels). They are reasonably comprehensive. 

#### Some issues ####
The fact that we couldn't write this in ST is a bit lame.. but I think we can get there. 
The css here is a mess. 
Adding and removing events causes this issue #97  (nothing to do with editor...)
The editor view is not really responsive. For example, it can send a `respond to [event]` message but it is not aware that the property changed to respond to that, so you have to [manually add it](https://github.com/UnitedLexCorp/SimpleTalk/compare/daniel-button-editor?expand=1#diff-d460e66340863153b40b711a201ef3dea6540293e5d5bc796755e4f238f0a473R320). This is easy enough here, but we know in more complicated situations the fact that there is no responsiveness between the target model and it's editor. Potentially having the editor add itself to the target and then using the `target.onPropChange()` could work. But then we'd have `editor.target` and `target.editor` attributes floating around... or add a model for the editor. This is the reason I left a [basic version](https://github.com/UnitedLexCorp/SimpleTalk/compare/daniel-button-editor?expand=1#diff-5c04e967413aec14e8cd22a4fe1f520620357e5acb7c9e24d4c5eb879fc22674) of that in (at first I had an editor model/view setup). TBD... 

Give it a go...

### DIFF
--- a/js/objects/views/editors/ButtonEditorView.js
+++ b/js/objects/views/editors/ButtonEditorView.js
@@ -1,4 +1,6 @@
-import PartView from '../PartView.js';
+import ColorWheelWidget from '../drawing/ColorWheelWidget.js';
+
+const closeIcon = `<svg xmlns="http://www.w3.org/2000/svg" class="icon icon-tabler icon-tabler-square-x" width="24" height="24" viewBox="0 0 24 24" stroke-width="2" stroke="currentColor" fill="none" stroke-linecap="round" stroke-linejoin="round"><path stroke="none" d="M0 0h24v24H0z" fill="none"></path><rect x="4" y="4" width="16" height="16" rx="2"></rect><path d="M10 10l4 4m0 -4l-4 4"></path></svg>`;
 
 const templateString = `
 <style>
@@ -61,18 +63,16 @@ const templateString = `
 <div class="editor-main">
     <input class="name"></input>
     <button class="script">Script</button>
-    <button class="background-color">Background Color</button>
-    <button class="font-color">Font Color</button>
-    <div class="events">
-        <input type="checkbox" id="click" name="click" checked>
-        <label for="click">click</label>
-        <input type="checkbox" id="mouseenter" name="mouseenter">
-        <label for="mouseenter">mouseenter</label>
-        <input type="checkbox" id="mouseover" name="mouseover">
-        <label for="mouseenter">mouseover</label>
+    <button class="background-color" name="background-color">Background Color</button>
+    <button class="font-color" name="font-color">Font Color</button>
+    <div class="events-display">
+        <input class="events"></input>
+        <div class="event-list"></div>
     </div>
 </div>
 `;
+
+const SVGParser = new DOMParser();
 
 class ButtonEditorView extends HTMLElement {
     constructor(){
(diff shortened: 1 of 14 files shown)


js/objects/views/editors/ButtonEditorView.js
```diff
@@ -0,0 +1,385 @@
...
+        // TODO: this is a hack b/c the editor is not properly responsive at the moment
+        let eventEl = eventListDiv.querySelector(`#${eventName}`);
+        if(eventEl){
+            return;
+        }
+        eventEl = document.createElement("div");
+        let eventSpan = document.createElement("span");
+        let closeSpan = document.createElement("span");
+        let xmlDocument = SVGParser.parseFromString(closeIcon, 'application/xml');
+        let closeSvgEl = xmlDocument.documentElement;
```
darth-cheney, line 363: Where does this get used/displayed?

darth-cheney (commented):
This looks good to me. Just a couple of points/issues:
1. Changing the button's text color also changes the color of the icons in the open halo, as in this screenshot: 
Screenshot from 2021-01-28 12-23-40
  
2. I can no longer adjust the size of the script editor window using the bottom right corner. This seems to trigger strange dragging behavior
  
Once we get some more of the critical parts of the system down, we should come back and explore the buttons in this editor. I was confused at first about how to remove an event listener for a button (the - sign is very small and the only clickable part of it). We might also consider an autocompletion dropdown for adding the event listeners which says the events available for the given Part.

js/objects/views/editors/ButtonEditorView.js
dkrasner: ah that's crud - will remove

dkrasner:
> This looks good to me. Just a couple of points/issues:
> 
>     1. Changing the button's text color also changes the color of the icons in the open halo, as in this screenshot:
>        Screenshot from 2021-01-28 12-23-40
> 

yeah i noticed that, but then like it but can remove; it's just a css thing on the halo which need to be explicit i think..


>     2. I can no longer adjust the size of the script editor window using the bottom right corner. This seems to trigger strange dragging behavior
ok i'll look into it, but i think this is already a problem in `master`  in which case i'll fix it in another PR

> 
> Once we get some more of the critical parts of the system down, we should come back and explore the buttons in this editor. I was confused at first about how to remove an event listener for a button (the - sign is very small and the only clickable part of it). We might also consider an autocompletion dropdown for adding the event listeners which says the events available for the given Part.
yeah it's shitty, i wasn't sure how we want to lay these out or whether we will the fact that the editor is not one thing but more an interface to open various sub-editors. If we like the latter, maybe we can just open an `events` sub-editor and have that nicely styled etc

PR #99, opened 2021-01-28
Daniel syntax debug

dkrasner:
### Main Points ###

This PR introduces a prototype [error handler](https://github.com/UnitedLexCorp/SimpleTalk/compare/daniel-syntax-debug?expand=1#diff-ad019d33cd101d6adeceb9553b2dfeeef98a8a808209c536107b05d948ff9ca2) 

The only types of errors it handles at the moment are `GrammarMatchError` and the rest it sends to System. But we like this setup it's clear how to add more cases. 

In the `GrammarMatchError` message there is now a [bit more](https://github.com/UnitedLexCorp/SimpleTalk/compare/daniel-syntax-debug?expand=1#diff-27807be337857d016752d52609f4312048ef006984f0d2e108676823ffe892f6R318) data coming from Ohm, which is then used by the error handler to mark up the script. 

All it does is add a `<<<` marker at the end of the guilty line and some additional information (if Ohm provides it). Moreover, it also opens the grammar and scrolls to the corresponding rule line (if Ohm provided that). Like this: 

image

If the script editor is not open - for example if the script is set programmatically - the error handler will open it. 

### Tests ###
I wrote basic tests, but excluded testing for the presence of the Grammar field display (maybe we don't want this). 

### Issues ###
This does not handle runtime errors, such as  `DoesNotUnderstand` command, but if we like this I think handling that in a similar way will be good. This is also one reason I wanted to open the script editor if it's not present, i.e. have a system command handy for it. For example, if you press a button and a DNU message is throw, I imagine the error handler will open the button editor at the corresponding line. 

Is our grammar too lax.  For example:
```
on click
    what
end click
```
does not throw a parse error..  Yes we can handle it at runtime but maybe something so glaring should fail parsing.

### DIFF
--- a/js/objects/System.js
+++ b/js/objects/System.js
@@ -300,7 +300,7 @@ const System = {
         if(!targetObject || targetObject == undefined){
             throw new Error(`System could not compile target object ${aMessage.targetId}`);
         }
-        
+
 
         // Attempt to parse the incoming SimpleTalk script string.
         // If there are grammatical errors, report them and bail.
@@ -309,10 +309,14 @@ const System = {
         // level of the script, which will create the JS handler functions
         let parsedScript = languageGrammar.match(aMessage.codeString);
         if(parsedScript.failed()){
+            // consider using the parse data from trace
+            // example: let tracedScript = languageGrammar.trace(aMessage.codeString);
+            // let tree = tracedScript.toString();
             let msg = {
                 type: "error",
                 name: "GrammarMatchError",
-                message: parsedScript.message,
+                parsedScript: parsedScript,
+                partId: aMessage.targetId
             };
             targetObject.sendMessage(msg, targetObject);
         } else {
@@ -329,8 +333,8 @@ const System = {
             );
             targetObject._semantics(parsedScript).interpret();
         }
-        
-        
+
+
         // Be sure to then update the
         // serialization for the target
         // part, thus adding the script to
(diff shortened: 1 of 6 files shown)


darth-cheney (commented):
Ok this is working well and looks good to me!
  
Concerning this comment you made:
> Is our grammar too lax. For example
```
on click
    what
end click
```
> does not throw a parse error.. Yes we can handle it at runtime but maybe something so glaring should fail parsing.'
  
This is not too lax nor should it throw any kind of parse error. It fits in with the design we have, which is that users can define arbitrary message names that are not any of the reserved keywords. Because we don't use `what` as a reserved keyword anywhere in the language, this will be interpreted as a message that is to be sent. Reacting to the DNU, however, is another matter that you rightly say we need to deal with.

dkrasner:
> Ok this is working well and looks good to me!
> 
> Concerning this comment you made:
> 
> > Is our grammar too lax. For example
> 
> ```
> on click
>     what
> end click
> ```
> 
> > does not throw a parse error.. Yes we can handle it at runtime but maybe something so glaring should fail parsing.'
> 
> This is not too lax nor should it throw any kind of parse error. It fits in with the design we have, which is that users can define arbitrary message names that are not any of the reserved keywords. Because we don't use `what` as a reserved keyword anywhere in the language, this will be interpreted as a message that is to be sent. Reacting to the DNU, however, is another matter that you rightly say we need to deal with.

yeah that's right; ok my next step then will be to handle runtime errors, but then I am thinking to implement a global handler/message name store which would do two things:

* allow for hinting and auto-completion on any given model, i.e. it would be be able to tell you what handlers are available for the given part and for parts up the stack (all the way System)
* allow some sort of a visual marker in both a script and the view, like a red border, whose model/parts do not handle all the messages listed. For example, you could write on a button
```
on click
   nonExistentHandler
end click
```
and press "save." No grammar error would be throw but `nonExistentHandler` would get some markup, while the button itself would get a red border until `nonExistenHandler` is added to the stack.

PR #100, opened 2021-01-28
Hand detection

ApproximateIdentity:
This integrates the hand detection stuff as global message handlers in system. This isn't complete, but it matches roughly what me and @darth-cheney discussed yesterday (at least my interpretation). Here's the flow:

Start up the video by running

```
    System.receiveMessage({
        type: "command",
        commandName: "startVideo",
        args: []
    });
```

Next instantiate the hand detection model by running

```
    System.receiveMessage({
        type: "command",
        commandName: "startHandDetectionModel",
        args: []
    });
```

Next run a hand detection on the video by running (logs to console)

```
    System.receiveMessage({
        type: "command",
        commandName: "detectHands",
        args: []
    });
```

There are three commits here that implement this that are really pretty simple so maybe look at those if you have questions.

### DIFF
--- a/js/objects/System.js
+++ b/js/objects/System.js
@@ -34,6 +34,7 @@ import ButtonEditorView from './views/editors/ButtonEditorView.js';
 import ohm from 'ohm-js';
 import interpreterSemantics from '../ohm/interpreter-semantics.js';
 
+const video = document.createElement('video');
 
 const System = {
     name: "System",
@@ -1440,6 +1441,28 @@ System._commandHandlers['saveHTML'] = function(senders){
     anchor.click();
     window.URL.revokeObjectURL(url);
     anchor.parentElement.removeChild(anchor);
+};
+
+System._commandHandlers['startVideo'] = () => {
+    if (video.srcObject !== null) {
+        return;
+    }
+    navigator.mediaDevices.getUserMedia({ video: true }).then(stream => {
+        video.srcObject = stream;
+        video.play();
+    });
+};
+
+System._commandHandlers['stopVideo'] = () => {
+    if (video.srcObject === null) {
+        return;
+    }
+    video.pause();
+    const tracks = video.srcObject.getTracks();
+    for (var i = 0; i < tracks.length; i++) {
+        tracks[i].stop();
+    }
+    video.srcObject = null;
 };
 
 /** Register the initial set of parts in the system **/
--- a/js/objects/System.js
+++ b/js/objects/System.js
@@ -355,10 +355,6 @@ const System = {
         let handler = this._commandHandlers[aMessage.commandName];
         if(handler){
             let boundHandler = handler.bind(this);
-            let originalSender;
-            if(aMessage.senders){
-                originalSender = this.partsById[aMessage.senders[0].id];
-            }
             return boundHandler(aMessage.senders, ...aMessage.args);
         } else {
             return this.doesNotUnderstand(aMessage);
--- a/js/objects/System.js
+++ b/js/objects/System.js
@@ -149,6 +149,7 @@ const System = {
         );
         worldView.setModel(worldModel);
         document.body.appendChild(worldView);
+        document.body.appendChild(video);
         document.body.appendChild(canvas);
 
 
--- a/js/objects/System.js
+++ b/js/objects/System.js
@@ -1512,6 +1512,58 @@ System._commandHandlers['stopHandDetectionModel'] = () => {
     unloadHandDetectionModel();
 };
 
+const scaleDim = (dim) => {
+    const scale = 0.7;
+    const stride = 16;
+    const evenRes = dim * scale - 1;
+    return evenRes - (evenRes % stride) + 1;
+};
+
+const detectHands = async () => {
+    const canvas = document.createElement('canvas');
+    canvas.width = video.videoWidth;
+    canvas.height = video.videoHeight;
+    const ctx = canvas.getContext('2d');
+    ctx.drawImage(video, 0, 0, canvas.width, canvas.height);
+    const scaledWidth = scaleDim(canvas.width);
+    const scaledHeight = scaleDim(canvas.height);
+    const image = tf.fromPixels(canvas).resizeBilinear([scaledHeight, scaledWidth]).expandDims(0);
+    const bboxes = await handDetectionModel.executeAsync(image).then(result => {
+        const [scores, boxes] = result;
+        const indices = tf.image.nonMaxSuppression(
+            boxes.reshape([boxes.shape[1], boxes.shape[3]]),
+            scores.reshape([scores.shape[1]]),
+            20,
+            0.5,
+            0.80).dataSync();
+        var bboxes = [];
+        var idx;
+        for (let i = 0; i < indices.length; i++) {
+            idx = indices[i];
+            var score = scores.get(0, idx, 0);
+            // Original order is [minY, minX, maxY, maxX] so we reorder.
+            var box = [
+                boxes.get(0, idx, 0, 1) * canvas.width, // minX
+                boxes.get(0, idx, 0, 0) * canvas.height, // minY
+                boxes.get(0, idx, 0, 3) * canvas.width, // maxX
+                boxes.get(0, idx, 0, 2) * canvas.height // maxY
+            ];
+            bboxes.push([score, box]);
+        }
+        return bboxes;
+    });
+    console.log(bboxes);
+};
+
+System._commandHandlers['detectHands'] = () => {
+    if (handDetectionModel === null) {
+        console.log("Error: no hand detection model loaded");
+        return;
+    }
+    detectHands();
+}
+
+
 /** Register the initial set of parts in the system **/
 System.registerPart('card', Card);
 System.registerPart('stack', Stack);
--- a/js/objects/System.js
+++ b/js/objects/System.js
@@ -35,6 +35,7 @@ import ohm from 'ohm-js';
 import interpreterSemantics from '../ohm/interpreter-semantics.js';
 
 const video = document.createElement('video');
+var handDetectionModel = null;
 
 const System = {
     name: "System",
@@ -1463,6 +1464,52 @@ System._commandHandlers['stopVideo'] = () => {
         tracks[i].stop();
     }
     video.srcObject = null;
+};
+
+// https://aaronsmith.online/easily-load-an-external-script-using-javascript/
+const loadScript = src => {
+    return new Promise((resolve, reject) => {
+        const script = document.createElement('script')
+        script.type = 'text/javascript'
+        script.onload = resolve
+        script.onerror = reject
+        script.src = src
+        document.head.append(script)
+    })
+}
+
+const loadHandDetectionModel = () => {
+    if (handDetectionModel === null) {
+        window.tf.loadFrozenModel(
+            "https://cdn.jsdelivr.net/npm/handtrackjs/models/web/ssdlitemobilenetv2/tensorflowjs_model.pb",
+            "https://cdn.jsdelivr.net/npm/handtrackjs/models/web/ssdlitemobilenetv2/weights_manifest.json"
+        ).then(model => {
+            console.log("hand detection model loaded");
+            handDetectionModel = model;
+        }).catch(err => {
+            console.log("error loading hand detection model");
+            console.log(err);
+        });
+    }
+}
+
+System._commandHandlers['startHandDetectionModel'] = () => {
+    if (typeof window.tf === 'undefined') {
+        loadScript("https://cdn.jsdelivr.net/npm/@tensorflow/tfjs@0.13.5/dist/tf.js").then(() => {
+            loadHandDetectionModel();
+        });
+    } else {
+        loadHandDetectionModel();
+    }
+};
+
+const unloadHandDetectionModel = () => {
+    console.log(handDetectionModel);
+    handDetectionModel = null;
+}
+
+System._commandHandlers['stopHandDetectionModel'] = () => {
+    unloadHandDetectionModel();
 };
 
 /** Register the initial set of parts in the system **/
--- a/js/objects/System.js
+++ b/js/objects/System.js
@@ -149,9 +149,10 @@ const System = {
         );
         worldView.setModel(worldModel);
         document.body.appendChild(worldView);
-        document.body.appendChild(video);
-        document.body.appendChild(canvas);
 
+        // Uncomment the following to see hand detection frames.
+        //document.body.appendChild(video);
+        //document.body.appendChild(canvas);
 
         // Create an initial blank Stack in this
         // case
--- a/js/objects/System.js
+++ b/js/objects/System.js
@@ -1555,24 +1555,26 @@ const detectHands = async (recipientId) => {
         console.log(bboxes);
     } else {
         let recipient = System.partsById[recipientId];
-        System.sendMessage({
+        let msg = {
             type: 'command',
             commandName: 'detectedHands',
             args: [JSON.stringify(bboxes, null, 4)]
-        },
-        System,
-        recipient);
+        };
+        System.sendMessage(msg, System, recipient);
     }
 };
 
-System._commandHandlers['detectHands'] = (senders, ...args) => {
+System._commandHandlers['detectHands'] = (senders, recipientId) => {
     if (handDetectionModel === null) {
         console.log("Error: no hand detection model loaded");
         return;
     }
-    let recipientId = null;
-    if (senders.length) {
-        recipientId = senders[0].id;
+    if (recipientId === undefined) {
+        if (senders !== undefined && senders.length) {
+            recipientId = senders[0].id;
+        } else {
+            recipientId = null;
+        }
     }
     detectHands(recipientId);
 }
--- a/js/objects/System.js
+++ b/js/objects/System.js
@@ -1541,13 +1541,11 @@ const detectHands = async (recipientId) => {
             idx = indices[i];
             var score = scores.get(0, idx, 0);
             // Original order is [minY, minX, maxY, maxX] so we reorder.
-            var box = [
-                boxes.get(0, idx, 0, 1) * canvas.width, // minX
-                boxes.get(0, idx, 0, 0) * canvas.height, // minY
-                boxes.get(0, idx, 0, 3) * canvas.width, // maxX
-                boxes.get(0, idx, 0, 2) * canvas.height // maxY
-            ];
-            bboxes.push([score, box]);
+            var box = {
+                upperLeft: [boxes.get(0, idx, 0, 1), boxes.get(0, idx, 0, 0)],
+                lowerRight: [boxes.get(0, idx, 0, 3), boxes.get(0, idx, 0, 2)]
+            };
+            bboxes.push({score: score, box: box});
         }
         return bboxes;
     });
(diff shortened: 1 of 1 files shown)


dkrasner:
@ApproximateIdentity I see the camera turn on when i run 
```
   System.receiveMessage({
        type: "command",
        commandName: "startVideo",
        args: []
    });
```

am i supposed to see anything else, as the messages are sent? (in console or otherwise). At the moment in console all I see is "hand detection model loaded" but nothing happens.

js/objects/System.js
```diff
@@ -1336,6 +1338,126 @@ System._commandHandlers['saveHTML'] = function(senders){
...
+        const script = document.createElement('script')
+        script.type = 'text/javascript'
+        script.onload = resolve
+        script.onerror = reject
+        script.src = src
+        document.head.append(script)
+    })
+}
+
+const loadHandDetectionModel = () => {
```
dkrasner, line 1375: maybe we should move all this into `utils/video.js` file. System needs to be cleaned up anyhow ...

ApproximateIdentity:
> @ApproximateIdentity I see the camera turn on when i run
> 
> ```
>    System.receiveMessage({
>         type: "command",
>         commandName: "startVideo",
>         args: []
>     });
> ```
> 
> am i supposed to see anything else, as the messages are sent? (in console or otherwise). At the moment in console all I see is "hand detection model loaded" but nothing happens.
> 
> I would also add some tests for these messages, at least to check for the presence of elements if any.

The "hand detection model loaded" console log is really more of a debugging thing right now. Maybe there should be a little light produced somewhere on the main page that goes green or something to show that it's ready, but I wasn't totally sure. That is in theory unimportant (of course from a UX perspective it's very important, but the actual coordinates for hands are the crucial part). I just wasn't sure about this presentation part of the stuff so figured it's best to throw it in as a simple version to make sure it's even in the right place (i.e. in system like I said).

But once everything is loaded, if you send the detectHand message, it will tell you the coordinates of any hands that it finds. That too is currently just logged to the console for debugging, but that's really just because I wasn't sure exactly the best way to send the message back. Either way this of course has to change since clearly logging the console for this isn't okay in any reasonable design.

js/objects/System.js
ApproximateIdentity: Yeah Eric floated basically the same idea. Maybe even have a standardized plugins folder. I just decided to dump it straight in here since this case is pretty simple, but I think before actually merging this, that (as well as the console logging and other stuff that this is missing) should be fixed. My first question was just if having objects in system responding to mesages in roughly this way is the way that seems to fit with everyone's ideas.

js/objects/System.js
```diff
@@ -1406,6 +1406,58 @@ System._commandHandlers['stopHandDetectionModel'] = () => {
...
+                boxes.get(0, idx, 0, 1) * canvas.width, // minX
+                boxes.get(0, idx, 0, 0) * canvas.height, // minY
+                boxes.get(0, idx, 0, 3) * canvas.width, // maxX
+                boxes.get(0, idx, 0, 2) * canvas.height // maxY
+            ];
+            bboxes.push([score, box]);
+        }
+        return bboxes;
+    });
+    console.log(bboxes);
```
ApproximateIdentity, line 1449: @darth-cheney @dkrasner 

This is the place were the hands that are detected are logged. Obviously this isn't right, but what do you think the best way to do this and have a sane design?

js/objects/System.js
darth-cheney: We definitely want to think about a design for System that is modular -- one that can accept plugins/mixins/whatever. For example, built-in System command handlers should probably be in some separate file(s). Vision is another example of a plugin. We can imagine threaded/remote interfaces also being plugins. We just need to think about it for a bit and map out some possible designs.

dkrasner: yeah for now we can just put this in utils, then deal with modules

dkrasner:
> > @ApproximateIdentity I see the camera turn on when i run
> > ```
> >    System.receiveMessage({
> >         type: "command",
> >         commandName: "startVideo",
> >         args: []
> >     });
> > ```
> > 
> > 
> > am i supposed to see anything else, as the messages are sent? (in console or otherwise). At the moment in console all I see is "hand detection model loaded" but nothing happens.
> > I would also add some tests for these messages, at least to check for the presence of elements if any.
> 
> The "hand detection model loaded" console log is really more of a debugging thing right now. Maybe there should be a little light produced somewhere on the main page that goes green or something to show that it's ready, but I wasn't totally sure. That is in theory unimportant (of course from a UX perspective it's very important, but the actual coordinates for hands are the crucial part). I just wasn't sure about this presentation part of the stuff so figured it's best to throw it in as a simple version to make sure it's even in the right place (i.e. in system like I said).
> 
> But once everything is loaded, if you send the detectHand message, it will tell you the coordinates of any hands that it finds. That too is currently just logged to the console for debugging, but that's really just because I wasn't sure exactly the best way to send the message back. Either way this of course has to change since clearly logging the console for this isn't okay in any reasonable design.


so I am not seeing anything here, after I run 
```System.receiveMessage({
        type: "command",
        commandName: "detectHands",
        args: []
    });
```

nothing happens

can you send me a screenshot of what i ought to expect

ApproximateIdentity:
> > > @ApproximateIdentity I see the camera turn on when i run
> > > ```
> > >    System.receiveMessage({
> > >         type: "command",
> > >         commandName: "startVideo",
> > >         args: []
> > >     });
> > > ```
> > > 
> > > 
> > > am i supposed to see anything else, as the messages are sent? (in console or otherwise). At the moment in console all I see is "hand detection model loaded" but nothing happens.
> > > I would also add some tests for these messages, at least to check for the presence of elements if any.
> > 
> > 
> > The "hand detection model loaded" console log is really more of a debugging thing right now. Maybe there should be a little light produced somewhere on the main page that goes green or something to show that it's ready, but I wasn't totally sure. That is in theory unimportant (of course from a UX perspective it's very important, but the actual coordinates for hands are the crucial part). I just wasn't sure about this presentation part of the stuff so figured it's best to throw it in as a simple version to make sure it's even in the right place (i.e. in system like I said).
> > But once everything is loaded, if you send the detectHand message, it will tell you the coordinates of any hands that it finds. That too is currently just logged to the console for debugging, but that's really just because I wasn't sure exactly the best way to send the message back. Either way this of course has to change since clearly logging the console for this isn't okay in any reasonable design.
> 
> so I am not seeing anything here, after I run
> 
> ```
>         type: "command",
>         commandName: "detectHands",
>         args: []
>     });
> ```
> 
> nothing happens
> 
> can you send me a screenshot of what i ought to expect

Oh I see I misunderstood. All you should see is a console log with an array of boxes. If it doesn't detect any hands, it should log an empty array. If it detects hands, then each hand will itself be logged as a score saying how much it thinks it's a hand and another array with the coordinates of the corners of the box. But if you're getting nothing logged at all, then the only real thing that makes sense is that I must be making some sort of assumption that fits my system, but doesn't fits yours. I'm gonna look through the code and figure out what I need to do to reproduce your side of things. I'll ping when I'm ready for you to try something.

But just to be clear, you do get the "model is loaded" console right? I.e. the only actual problem here is the detection?

ApproximateIdentity:
@dkrasner The latest version I pushed makes the canvas visible (at the bottom of the screen). Now when you send the detectHands message, the canvas should take the screenshot on which hand detection is done. If you send the detectHands message does it actually put an image in that box? If it does, can you right-click save as and send me the image so I can do a little debugging on my end?

dkrasner:
> > > > @ApproximateIdentity I see the camera turn on when i run
> > > > ```
> > > >    System.receiveMessage({
> > > >         type: "command",
> > > >         commandName: "startVideo",
> > > >         args: []
> > > >     });
> > > > ```
> > > > 
> > > > 
> > > > am i supposed to see anything else, as the messages are sent? (in console or otherwise). At the moment in console all I see is "hand detection model loaded" but nothing happens.
> > > > I would also add some tests for these messages, at least to check for the presence of elements if any.
> > > 
> > > 
> > > The "hand detection model loaded" console log is really more of a debugging thing right now. Maybe there should be a little light produced somewhere on the main page that goes green or something to show that it's ready, but I wasn't totally sure. That is in theory unimportant (of course from a UX perspective it's very important, but the actual coordinates for hands are the crucial part). I just wasn't sure about this presentation part of the stuff so figured it's best to throw it in as a simple version to make sure it's even in the right place (i.e. in system like I said).
> > > But once everything is loaded, if you send the detectHand message, it will tell you the coordinates of any hands that it finds. That too is currently just logged to the console for debugging, but that's really just because I wasn't sure exactly the best way to send the message back. Either way this of course has to change since clearly logging the console for this isn't okay in any reasonable design.
> > 
> > 
> > so I am not seeing anything here, after I run
> > ```
> >         type: "command",
> >         commandName: "detectHands",
> >         args: []
> >     });
> > ```
> > 
> > 
> > nothing happens
> > can you send me a screenshot of what i ought to expect
> 
> Oh I see I misunderstood. All you should see is a console log with an array of boxes. If it doesn't detect any hands, it should log an empty array. If it detects hands, then each hand will itself be logged as a score saying how much it thinks it's a hand and another array with the coordinates of the corners of the box. But if you're getting nothing logged at all, then the only real thing that makes sense is that I must be making some sort of assumption that fits my system, but doesn't fits yours. I'm gonna look through the code and figure out what I need to do to reproduce your side of things. I'll ping when I'm ready for you to try something.
> 
> But just to be clear, you do get the "model is loaded" console right? I.e. the only actual problem here is the detection?
yeah is see the console "model is loaded" and I know the camera is streaming, since I need to authorize that

but then i don't see anything else, nothing in console

ApproximateIdentity:
No that's actually correct there. I've just been really unclear. That final line with the empty array is the result due to it probably not detecting any hands. So if you put a hand in from of the screen itself and send the message it would hopefully detect it correctly by displaying an array of length 1. Does that make sense?

That array result is what I need to return to the requester somehow, but I wasn't totally sure the best way.

dkrasner:
yeah i tried putting my hands in front but nothing happened

ApproximateIdentity:
Here's the first picture showing the log after sending the three messages. You can see that my array isn't empty:

pic1

Here's the same array expanded:

pic2

Here's the main window scrolled down so that the image used in the detection shows up in hacked in canvas I put there (in the final commit I pushed to the branch). If I were to use the pixel values I got above, they would form a box around my hand.

pic3

Are you at least seeing a picture on the main screen like my last picture example?

dkrasner:
yeah i am seeing the video, but it's always a frozen frame, and i get that empty array and nothing further no matter what i do 

image


(I am on FF btw)

ApproximateIdentity:
Okay I pushed one more change that should make things clearer. In my original version (i.e. 2 commits back) neither the video nor the canvas were visible. Then the second to last commit made the canvas visible while the last commit that I just pushed makes the video itself visible. So now when you send the detectHands message it should cause the canvas frame to the right of the video to freeze frame and that is the frame that detection is done on. So basically at this point it should be easier to test things. The question we have with your setup is: no matter how how you move your hands in the frame does it _never_ pick up your hands? That would point to a modeling problem.

Sorry for making all this stuff hidden. I think it makes sense to have it hidden _if_ it works, but apparently I put too much faith in my code actually working.

Here's a picture from my machine to make it a little clearer what's happening:

darth-cheney (commented):
Aside from my inline comments, this is working for me. I even got Sabriya to push return the send the message while I had both hands up, and I got two arrays. Seems to work in that sense.

js/objects/System.js
ApproximateIdentity: That sounds like a good idea in principle. At least I think the design is good at a high-level. I'm a little worried about performance though. We'll have to test things on various machines, but it seems hard to get more than 15-20 frames a second for the hand recognition. So then the question becomes how much of the overall system throughput is eaten up by that message stream. Maybe it doesn't matter too much, but we'll have to play around with it to see if we run into any issues. Either way I think doing it this way or something similar is right since it is a clean design. Can mess with it afterwards if it has performance issues.

ApproximateIdentity:
> Aside from my inline comments, this is working for me. I even got Sabriya to push return the send the message while I had both hands up, and I got two arrays. Seems to work in that sense.

Yeah I took away the "fire repeately mode" since I wasn't totally sure how to fit it into the message passing stuff, but in hindsight I think forcing you to click it like that every time is extremely confusing (especially given that I was hiding the video and canvas elements). So it made for no real feedback. I guess the first question though is why @dkrasner isn't getting something back ever.

Daniel if you're still seeing nothing can you maybe find a couple frames that have a really clear hand in them (usually it doesn't need to be that clear, but a little clear is good) and then send them to me? The reason I wonder is if your camera might be (say) higher resolution or something and that is causing an issue. At least this way I should be able to reproduce the image that you have and run it through the modeling on my end and see if it detects a hand here.

dkrasner:
@ApproximateIdentity i can't get to a any frame past (what looks to be) the initial one, ie when the detect message is sent

dkrasner:
> Aside from my inline comments, this is working for me. I even got Sabriya to push return the send the message while I had both hands up, and I got two arrays. Seems to work in that sense.

perhaps not custom DOM elements, but custom events such as "handmove" or "handdetect" or "handbind" etc 

then these can be fired by system, or a system module which handles data coming from the API, and listened for by specific part/model views

dkrasner:
@ApproximateIdentity i don't want to block you from merging this in. So if you feel like it's ready and @darth-cheney is ok with it just merge

ApproximateIdentity:
Well I was looking into hooking this up to the events work you did as Eric suggested, but I'm a little confused as to the best way to do so. Does it make sense to have it there? I'm not sure it makes sense to send an event stream that's being dropped if there's not listeners since the computation itself is extremely heavy. Right now we can't do the detection at full frame-rate meaning it would be running at full blast only to be dropped if I did this in the stupidest way possible.

But I could also merge this as is if it's not totally clear how to do this. Other than the fact that the message is currently just logged to the console (which is pointless), it is functional as a "pull only" event stream. So basically if the messages were sent back, they could be interpreted that way. @darth-cheney what do you think? (If what I'm saying makes sense.)

dkrasner:
Whenever the peripheral (hand, ball) is locked onto an object in SimpleTalk (cursor, button...) the even message stream has to be as granular as a general mouse move event stream. 

So the dynamic i see is, you will have the option to lock in an object **from** SimpleTalk. For example, you can add a button to Halo which will then send the messages above for a specific part (with id `partID`), start the video, load the model and start tracing. Then the messages coming back with the moveX, moveY data will look something like

```
msg = {
type: "event",
eventName: "peripheralMove",
targetId: partId,
data: [moveX, moveY]
}
```


all we'll need is a little middleware which will take this message and actually send it to the part in question `part.sendMessage(msg, part)`  (which means it will have to find the part with that id `partId`).

ApproximateIdentity:
Okay the last commit implements a basic "send hand detected boxes back" message approach to this. To test it do the following:

1. Send the following two messages in the console to start up the video and model:

```
    System.receiveMessage({
        type: "command",
        commandName: "startVideo",
        args: []
    });

    System.receiveMessage({
        type: "command",
        commandName: "startHandDetectionModel",
        args: []
    });
```

2. Click the "Add button to Card" button and change its script to be:

```
on click
	detectHands
end click

on detectedHands bboxes
	answer bboxes
end detectedHands
```
3. Hold your hands up and click the button to test things out. Here is a screenshot of what I see:

hand_detection

ApproximateIdentity:
@darth-cheney @dkrasner 

Actually the more I think about it, maybe this should just be merged? The reason I think so is because at this point it's technically functioning as a "video hand detection" setup. I think I should remove the two lower videos that are only there for debugging, but other than that this is technically functional.

Do you disagree?

ApproximateIdentity:
The last commit I pushed adds support for specifying the recipientId. It is an extremely similar flow to before (i.e. there are no changes to the model video and model instantiation), but now you can do the following. Create two buttons. I'll assume the second button's id is 14.

1. Add the following script to the first button:
```
on click
	detectHands 14
end click
```

2. Add the following script to the second button (i.e. the one with id 14):
```
on detectedHands bboxes
	answer bboxes
end detectedHands
```

Then it works as before. HEre are couple screenshots:





Note: If you don't specify a recipient it just sends it back to the sender (as before). If you don't specify a sender or a recipient (e.g. you send the message using System in the js console), then it just logs the answer to the console. This is really just a convenient for debugging.

ApproximateIdentity:
Okay so my last commit removed the pixel scaling. That way whatever using/presenting the results can decide what to do. It also cleaned up the format to be a bit more readable. Here is what a click looks like now:

better

ApproximateIdentity:
@darth-cheney @dkrasner 

The last two changes should allow a more reasonable programming model. Thoughts?

dkrasner:
yeah that looks good. 

Technically you are using the fact that our grammar is set up to take any grammatically correct looking statement line/command. But the implicit (should be explicit in the design specs), is that this is for SimpleTalk person defined commands, like you `on detectedHands` one. 

So I would add this to the grammar and interpreter (this PR or next...). That way you get object specifiers for free, built it, like [so](https://github.com/UnitedLexCorp/SimpleTalk/blob/master/js/ohm/simpletalk.ohm#L184) and you immediately write:

```
on click
    detectHands   -- meaning on this here button or whatever you are clicking on
end click
```

```
on click
    detectHands in part 14  -- by ID
end click
```
```
on click
    detectHands in button "cool button"  -- by name
end click
```

otherwise we are liable to be running two syntax grammars.  And this makes things easier to test in my view, since we have grammar tests and interpreter tests etc


On the design decision of what you actually send to the part. This is kind of interesting, but i would gravitate to uniformity with other peripherals. For example, do want to send `{type: "command", commandName: "mouseMove, args[X, Y]}`
or in the case of a mouse move do we want to send `{type: "command", commandName: "move, args[X, Y]}`?

We definitely want to have a command name "move," so you can programmatically move things around. And maybe we want these peripheral moves to then send the `move` command, or do we want to keep things simple and just stick with one...

ApproximateIdentity:
The grammar part definitely makes sense, but I think it probably does make sense to do that as a separate PR just to get this one in.

I can see the idea of having a move or a mouseMove command coming out, but at the moment the question is what to even send. If we wanted to send a single mouse move point we should send the average of these two upper left and corner points so we have the center. But if we wanted to use the whole box somehow we should be returning something like "draw box from [here] to [here]..." or something. I feel like in a way the current model piece is actually returning the correct thing (i.e. the detected boxes in current form) and if we wanted to have a separate e.g. mouseMove part or something then that should sit in between the requester and the model portion doing the transformations. I guess my main point is that there are just too many different possibilities at the moment and so I think as a first step it makes sense to keep it essentially in its current form (though it would be nice if we could skip the JSON.stringify() call) and then do the necessary interpretation in the parts. I think if there's a clearly better way it'll fall out in use anyway.

So basically I would personally just merge this now and then work on the grammar parts in a separate PR and then think about the move X, Y or mouseMove X, Y questions down the line. Thoughts about that?

dkrasner:
i think that's right

get this in
add to the grammar
experiment with what actually works best (what is the data from these boxes we want to use to move a button, a mouse cursor a drawing cursor - might not be the same and I figure we probably want to be able to toggle different options, could be in js code, and test what feels best)

ApproximateIdentity:
Okay cool. Then I'm merging this now and will then sort out the grammar side in the next PR and go from there.

PR #103, opened 2021-02-02
Image Part

darth-cheney:
## What ##
This PR implements a more capable replacement for the Svg part that can also handle the loading and display of binary images.
  
### Two Modes ###
The `Image` part essentially operates under the hood in one of two modes: as an SVG or as a binary image. If it is an SVG, the `imageData` part property will contain the SVG markup as text. If it is a binary, the `imageData` part property contains a data url base64 encoded string representing the mimetype and bytes of the image. Note that we also have a property called `mimeType` that is used when fetching image data from a remote or local source.
  
### Fetching Image Data ###
Fetching the remote image data is handled on the Part side via a private command handler called `loadImageFrom` which takes a url as its single argument. This can be sent to the image as a message, or invoked within the Image part's script.  The views send this message when the url has been changed via Halo options (see below).
  
Views listen for changes to the source and/or imageData and then respond accordingly. They do not do any remote fetching themselves. Binary imageData is an encoded data url, which is given to the shadow img tag, but this is copied from memory and not fetched (again).
  
### Halo Link ###
I have added a new custom button that allows the user to change the image source link. It will have no effect if you keep the existing url. This allows us to quickly and easily open remote images for testing.
  
### Resizing and Aspect Ratio ###
For the moment, this feature is *still broken*. I have added [some checks in Halo](https://github.com/UnitedLexCorp/SimpleTalk/blob/eric-image-part/js/objects/views/Halo.js#L255) for a new view property called `preserveAspectOnResize`. If this is set to true, we want to ensure that the aspect ratio is preserved when resizing via the Halo resizer. Right now we have an algorithm in that section that only sort of works, and we need to figure out a better solution.
  
Note that this is only true for binary images and not for SVGs.
  
#### Example To Test/Try ####
Open the bootstrap and add an image to the card. It will add an SVG default image. Open a halo on it and click the link button, then paste in this url: [https://images.unsplash.com/photo-1610907861125-45c5e76b5cc2?ixlib=rb-1.2.1&q=80&fm=jpg&crop=entropy&cs=tinysrgb&w=1080&fit=max](https://images.unsplash.com/photo-1610907861125-45c5e76b5cc2?ixlib=rb-1.2.1&q=80&fm=jpg&crop=entropy&cs=tinysrgb&w=1080&fit=max). You'll have to scroll down the page to get to the resizer, but try to resize to a smaller size and you'll see what our implementation currently does.


dkrasner:
cool this is good, code is good, tests are good

some issues:

it takes a while (maybe 5-10s) for the initial SVG to load, even though it's a local fetch. Not sure why. Basically you see nothing on the screen for a bit when you click the "add image" button and then the huge SVG appear. 
Regardless can we make the default smaller, like it was before.

Can we in general make sure that the images don't go beyond the viewport in size?

when Halo is enabled on an image it doesn't wrap it anymore, so it's not calculating the size of the part view properly

image

darth-cheney:
> it takes a while (maybe 5-10s) for the initial SVG to load, even though it's a local fetch. Not sure why. Basically you see nothing on the screen for a bit when you click the "add image" button and then the huge SVG appear.
Regardless can we make the default smaller, like it was before.
  
> when Halo is enabled on an image it doesn't wrap it anymore, so it's not calculating the size of the part view properly
  
Yikes. It looks like this all works fine in Chrome (which I was testing in this time) but not in FF. I'm going to track down the issue(s) this morning

darth-cheney:
Ok @ApproximateIdentity and @dkrasner  -- I have pushed a series of commits that should take care of all the things we talked about. Let me know if this is good enough and we can just merge it in.
  
@dkrasner there is some on again off again test failing regarding script grammar checking. At this point I'm not sure what the correct assumption is for the test.

ApproximateIdentity:
I think it looks good overall. It doesn't fix one of the things I wanted to fix which is that the little resizer box should not diverge from the mouse itself. But I think that's an issue with the halo implementation and not this (and hence a bit out of scope of this PR). But I do personally think it should be changed wherever the issue is located.

ApproximateIdentity:
To clarify on the last point: your "invariant" is that the resizer should stick to the lower right corner of the object and hug it exactly. I don't think that's right. I think the resizer should hug the mouse position and then the internal image should scale as large as it can while preserving its aspect ratio.

darth-cheney:
@ApproximateIdentity Gonna kick the can on the mouse tracking issue and come back to it once we've done a few passes on the overall CSS/Layouts.

PR #105, opened 2021-02-05
Object Specifiers v1

darth-cheney:
## What ##
This PR represents a first true pass at "Object Specifiers": pieces of SimpleTalk code that are like queries that resolve to a specific Part id.
  
Examples include:
```simpletalk
first field of this card
button 3 of second card of current stack
field "some named field" of card "some named card" of current stack
```
  
These are designed to be used in tandem with System or other commands that can take a Part id as an argument. They can also be used with the built in `set <propName> to <value> in <specifiedPart>` constructions.
  
### Grammar and Rules ###
The grammatical rules here are fairly straightforward. There are two general sub-types of Object Specifier: Terminal and Partial. A terminal part is one that you cannot query any deeper. These include `current card/stack`, `this <partName>` and any part referenced by id.
  
We introduce here the idea of a "QueriedSpecifier", which is either two QueriedSpecifiers together, or a PartialSpecifier ending in "of". In the `ObjectSpecifier` subrules, we use this to determine a full specification -- a query that ends in a terminal specifier or an implicit partial (see below).
  
We have also added numerical keywords "first" to "tenth" to the keyword list.
  
**NOTE**: We have changed the reference by id pattern to be `part id <num>` or `<partType> id <num>`.
  
### Semantics ###
The semantic implementation of ObjectSpecifier is a bit complex, but follows a consistent set of simple rules, which are the following:
1. Any `ObjectSpecifier` specific semantic action will return a Part id;
2. Any `PartialSpecifier`, `TerminalSpecifier`, or `QueriedSpecifier` action will return a function that takes a part as its context, and returns another part. This is used to chain lookups on the fly. You can see some of it in action  [here](https://github.com/UnitedLexCorp/SimpleTalk/blob/eric-object-specifier/js/ohm/interpreter-semantics.js#L715) and [here](https://github.com/UnitedLexCorp/SimpleTalk/blob/eric-object-specifier/js/ohm/interpreter-semantics.js#L691), for example.
  
#### `this card/stack` vs `current card/stack` ####
Note that we treat `current card` and `this card` differently -- the former refers to the current card that the user is seeing. The latter refers to the card that contains the part that is actually executing the script.
  
#### Implicit `this card` ####
When looking up a specifier like `field 1` without any terminal, we assume an implicit `of this card` on the end.
  
### Window and friends ###
This PR does not implement anything with consideration for the window part outside of generic capabilities of any other part. Windows are special types of containers though, so we will want to update these rules once we have a better idea of how we'd use them. I'm imagining that will come when we also implement a more generic layout area container, which will require the same considerations.
  
### Use of `this` and `current` ###
Grammatically we have an open question about how to use `current` inside ObjectSpecifiers. I have intentionally skipped [this test](https://github.com/UnitedLexCorp/SimpleTalk/blob/eric-object-specifier/js/ohm/tests/test-grammar.js#L505) because in my opinion it tests something that is potentially nonsense: what is a "current button," for example?
  
I have intentionally limited the `current` keyword to only deal with cards, stacks, and (though not used) backgrounds.
  
Similarly, we should decide how we want to deal with `this` when invoked inside a parts script, but with a part name that does not match that of the part context. For example, imagine I have a button script that contains the following line:
```simpletalk
-- The following is in a button, and
-- therefore doesn't really make sense
-- but is grammatically permitted
on click
    set "name" to "nonsense" in this field
end click
```
  
For the moment, I have [added a check](https://github.com/UnitedLexCorp/SimpleTalk/blob/eric-object-specifier/js/ohm/interpreter-semantics.js#L654) that throws an error if the specified part type does not match the type of the part that is executing the script.
The part context here is a button, not a field. What should happen? Should this even be allowed grammatically? Note also that `set "name" to "nonsense" in this button` is equivalent to just `set "name" to "nonsense"` (the part executing the statement is implicitly the context).
  
## Tests ##
We have two tests files, [one for the grammar](https://github.com/UnitedLexCorp/SimpleTalk/blob/eric-object-specifier/js/ohm/tests/test-object-specifier-grammar.js) and [one for the interpretation](https://github.com/UnitedLexCorp/SimpleTalk/blob/eric-object-specifier/js/objects/tests/test-object-speficier-with-preload.js). All of those tests are currently passing.
  
@dkrasner there is some Field error formatting thing that is not passing and I'm not sure what's supposed to be happening there. It's the only test in the suite not passing.


dkrasner:
> ## What
> 
> This PR represents a first true pass at "Object Specifiers": pieces of SimpleTalk code that are like queries that resolve to a specific Part id.
> 
> Examples include:
> 
> ```
> first field of this card
> button 3 of second card of current stack
> field "some named field" of card "some named card" of current stack
> ```
> 
> These are designed to be used in tandem with System or other commands that can take a Part id as an argument. They can also be used with the built in `set <propName> to <value> in <specifiedPart>` constructions.
> ## Implementation
> ### Grammar and Rules
> 
> The grammatical rules here are fairly straightforward. There are two general sub-types of Object Specifier: Terminal and Partial. A terminal part is one that you cannot query any deeper. These include `current card/stack`, `this <partName>` and any part referenced by id.
> 
> We introduce here the idea of a "QueriedSpecifier", which is either two QueriedSpecifiers together, or a PartialSpecifier ending in "of". In the `ObjectSpecifier` subrules, we use this to determine a full specification -- a query that ends in a terminal specifier or an implicit partial (see below).
> 
> We have also added numerical keywords "first" to "tenth" to the keyword list.
> 
> **NOTE**: We have changed the reference by id pattern to be `part id <num>` or `<partType> id <num>`.
> ### Semantics
> 
> The semantic implementation of ObjectSpecifier is a bit complex, but follows a consistent set of simple rules, which are the following:
> 
>     1. Any `ObjectSpecifier` specific semantic action will return a Part id;
> 
>     2. Any `PartialSpecifier`, `TerminalSpecifier`, or `QueriedSpecifier` action will return a function that takes a part as its context, and returns another part. This is used to chain lookups on the fly. You can see some of it in action  [here](https://github.com/UnitedLexCorp/SimpleTalk/blob/eric-object-specifier/js/ohm/interpreter-semantics.js#L715) and [here](https://github.com/UnitedLexCorp/SimpleTalk/blob/eric-object-specifier/js/ohm/interpreter-semantics.js#L691), for example.

can you add inline examples in the interpreter so it's easier to see right away what function is dealing with what type of statement
> 
> 
> #### `this card/stack` vs `current card/stack`
> 
> Note that we treat `current card` and `this card` differently -- the former refers to the current card that the user is seeing. The latter refers to the card that contains the part that is actually executing the script.
> #### Implicit `this card`
> 
> When looking up a specifier like `field 1` without any terminal, we assume an implicit `of this card` on the end.
> ## Caveats and Open Questions
> ### Window and friends
> 
> This PR does not implement anything with consideration for the window part outside of generic capabilities of any other part. Windows are special types of containers though, so we will want to update these rules once we have a better idea of how we'd use them. I'm imagining that will come when we also implement a more generic layout area container, which will require the same considerations.
> ### Use of `this` and `current`
> 
> Grammatically we have an open question about how to use `current` inside ObjectSpecifiers. I have intentionally skipped [this test](https://github.com/UnitedLexCorp/SimpleTalk/blob/eric-object-specifier/js/ohm/tests/test-grammar.js#L505) because in my opinion it tests something that is potentially nonsense: what is a "current button," for example?
i think it's fine to leave it in, as per the comment below
> 
> I have intentionally limited the `current` keyword to only deal with cards, stacks, and (though not used) backgrounds.
> 
background should not be included here, I am systematically removing all mentions of them in the core code
> Similarly, we should decide how we want to deal with `this` when invoked inside a parts script, but with a part name that does not match that of the part context. For example, imagine I have a button script that contains the following line:
> 
> ```
> -- The following is in a button, and
> -- therefore doesn't really make sense
> -- but is grammatically permitted
> on click
>     set "name" to "nonsense" in this field
> end click
> ```
> 
> For the moment, I have [added a check](https://github.com/UnitedLexCorp/SimpleTalk/blob/eric-object-specifier/js/ohm/interpreter-semantics.js#L654) that throws an error if the specified part type does not match the type of the part that is executing the script.
> The part context here is a button, not a field. What should happen? Should this even be allowed grammatically? Note also that `set "name" to "nonsense" in this button` is equivalent to just `set "name" to "nonsense"` (the part executing the statement is implicitly the context).
i  don't it should be grammar's role to restrict such specifiers. It is ok to say something that doesn't make sense, but if you want a specific response you should think about what you are saying. "The brown dog jump overs the lazy fox"... 

and, if we force this into the grammar, we'll be chasing all sorts of nonsense variants. `the 7th window of the 2nd button of ...` etc  So best to just throw a `ReferenceError` with as much info as possible and let ErrorHandler deal with it. 


> ## Tests
> 
> We have two tests files, [one for the grammar](https://github.com/UnitedLexCorp/SimpleTalk/blob/eric-object-specifier/js/ohm/tests/test-object-specifier-grammar.js) and [one for the interpretation](https://github.com/UnitedLexCorp/SimpleTalk/blob/eric-object-specifier/js/objects/tests/test-object-speficier-with-preload.js). All of those tests are currently passing.
> 
> @dkrasner there is some Field error formatting thing that is not passing and I'm not sure what's supposed to be happening there. It's the only test in the suite not passing.

don't worry about this, it is sorted in #104

js/objects/System.js
```diff
@@ -497,19 +497,11 @@ const System = {
     //     }
     // },
 
-    setProperty(senders, propName, value, objectId, objectType, context){
```
dkrasner, line 500: yeah this is good, get rid of this from System and all parts is the right way to go

js/objects/tests/#test-repeat-with-preload.js#
```diff
@@ -0,0 +1,304 @@
+import chai from 'chai';
```
dkrasner, line 1: is this is a stash file? looks like you don't want this here

js/objects/tests/test-object-speficier-with-preload.js
```diff
@@ -0,0 +1,341 @@
...
+        let partContext;
+
+        before(() => {
+            // We set the 'this card' part context to be the 2nd card in the stack, which
+            // is different from the current card (which is the 3rd card at this point).
+            // Remember that 'current card' and 'this card' can be different, if the script
+            // calling it is in a part that is not within the currently displayed card
+            partContext = window.System.getCurrentStackModel().subparts.filter(subpart => {
+                return subpart.type == 'card';
+            })[1];
```
dkrasner, line 238: do you want a check to make sure something stupid doesn't happen and the selected card is actually .current

js/ohm/interpreter-semantics.js
```diff
@@ -566,6 +545,177 @@ const createInterpreterSemantics = (partContext, systemContext) => {
...
+
+        ObjectSpecifier_compoundQueryWithTerminal: function(queriedSpecifier, terminalSpecifier){
+            // The terminal here is the ultimate part context
+            let finalPart = terminalSpecifier.interpret()();
+            let result = queriedSpecifier.interpret()(finalPart);
+            return result.id;
+        },
+
+        ObjectSpecifier_singleNonTerminal: function(partialSpecifier){
+            // A single non-terminal object specifier is one
```
dkrasner, line 705: it would be really helpful to have example scripts here. You can sniff out what it meant here from the comment and function name but it's not always 100% clear or unambiguous

js/ohm/interpreter-semantics.js
```diff
@@ -603,6 +753,33 @@ const createInterpreterSemantics = (partContext, systemContext) => {
             return result;
         },
 
+        numericalKeyword: function(numeralName){
```
dkrasner, line 756: do we want to allow a more general `Nth` option? which you can then just parse as `return string.split("th")[0]`

would be a single line but make our grammar/semantics more flexible

js/ohm/tests/test-object-specifier-grammar.js
```diff
@@ -0,0 +1,146 @@
...
+};
+
+const semanticMatchFailTest = (str, semanticType) => {
+    const typeMatch = g.match(str, semanticType);
+    assert.isFalse(typeMatch.succeeded());
+};
+
+const partTypes = [
+    "card",
+    "background",
```
dkrasner, line 25: remove

ApproximateIdentity:
@darth-cheney What's an example script that I can add to a button in the bootstrap-example page to make use of these changes? Sorry if this is trivial I'm always especially bad at the scripting stuff.

dkrasner:
@ApproximateIdentity 

add a few buttons, call them "Button 1" and "Button 2"

on button 1 write a script like

```
on click
    set "background-color" to "blue in button "Button 2" of current card
end click
```

on button 2 write a script like
```
on click
      respond to "mouseover"
end click

on mouseover
     set "font-color" to "red" in this button
end mouseover
```

(i didn't run these but they should work) 
on mouseover

PR #107, opened 2021-02-11
Scripted Drawing and Context Fixes

darth-cheney:
## What
This PR does two things:
1. Gives us the ability to execute basic drawing commands on the Drawing part via scripting/built-in command handlers;
2. Fixes an important issue regarding ExecutionContext, where old contexts were not being restored and variables were being overwritten.

## ExecutionContext ##
See the explanation in [this commit](https://github.com/UnitedLexCorp/SimpleTalk/commit/cd15af59d135e22af1f17a7854028d236859f918)
  
## Drawing Commands ##
We currently support the following commands: `beginDraw`, `finishDraw`, `lineTo`, `moveTo`, `stroke`. See the example script below.
  
What other drawing operations should we support? Should values like line width, color, fillstyle, etc be set as properties on the model, or should they be set via commands? The latter would work, since the model activates its own canvas context after `beginDraw` is executed.
  
Note also that for now the draw operations on the view side are not linked to these model side commands -- this is an optimization that (I believe) permits smoother mouse painting.
  
## Example ##
Here is an example script that works:
```simpletalk
on myDrawing
	put 50 into startX
	put 50 into startY
	put 200 into currentSize
	put 5 into increment
	beginDraw
	repeat until currentSize < 10
		drawSquare startX, startY, currentSize
		put (currentSize - (increment * 4)) into currentSize
		put (startX + increment) into startX
		put (startY + increment) into startY
	end repeat
	finishDraw
end myDrawing


on drawSquare x, y, size
	moveTo x, y
	lineTo (x + size), y
	lineTo (x + size), (y + size)
	lineTo x, (y + size)
	lineTo x, y
	stroke
end drawSquare
```
  
And produces: 
Screenshot from 2021-02-11 13-50-13

### DIFF
--- a/js/objects/parts/Drawing.js
+++ b/js/objects/parts/Drawing.js
@@ -55,11 +55,21 @@ class Drawing extends Part {
     }
 
     setupDrawingCommands(){
-        this._commandHandlers['lineTo'] = this.lineTo;
-        this._commandHandlers['moveTo'] = this.moveTo;
-        this._commandHandlers['beginDraw'] = this.beginDraw;
-        this._commandHandlers['endDraw'] = this.endDraw;
-        this._commandHandlers['stroke'] = this.stroke;
+        this.setPrivateCommandHandler('lineTo', (senders, ...args) => {
+            this.lineTo(...args);
+        });
+        this.setPrivateCommandHandler('moveTo', (senders, ...args) => {
+            this.moveTo(...args);
+        });
+        this.setPrivateCommandHandler('beginDraw', (senders, ...args) => {
+            this.beginDraw(...args);
+        });
+        this.setPrivateCommandHandler('finishDraw', (senders, ...args) => {
+            this.endDraw(...args);
+        });
+        this.setPrivateCommandHandler('stroke', (senders, ...args) => {
+            this.stroke(...args);
+        });
     }
 
     /* Scriptable Drawing Commands */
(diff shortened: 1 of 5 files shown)


js/objects/parts/Drawing.js
```diff
@@ -55,11 +55,22 @@ class Drawing extends Part {
     }
 
     setupDrawingCommands(){
-        this._commandHandlers['lineTo'] = this.lineTo;
-        this._commandHandlers['moveTo'] = this.moveTo;
-        this._commandHandlers['beginDraw'] = this.beginDraw;
-        this._commandHandlers['endDraw'] = this.endDraw;
-        this._commandHandlers['stroke'] = this.stroke;
+        this.setPrivateCommandHandler('lineTo', (senders, ...args) => {
+            console.log('is this working?');
```
dkrasner, line 59: remove

dkrasner:
so I think this is great!

one thing i wasn't sure of is how you got than drawing script to in the screenshot. I simply added an `on click` wrapper for `myDrawing` since `st-drawing` responds to clicks at the moment.

I did see this error in console though:

image



To answer your question "what else?" I would definitely add a "clear" command. This could be drawing specific, like `clearDrawing` but i would simply prefer something we can overload `clear [object view]` to which views that understand it can respond as needded

darth-cheney:
@dkrasner Gonna work on the small changes (and that console error) now, but I have opened an issue about the broader ExecutionContext issue here: https://github.com/UnitedLexCorp/SimpleTalk/issues/109
  
I say we merge this in just for the Drawing stuff and I will revisit the context issue immediately after in a separate branch

PR #108, opened 2021-02-12
css refactoring

dkrasner:
### Main Points ###

Here we refactor and cleanup the css, which includes the [core.css](https://github.com/UnitedLexCorp/SimpleTalk/compare/daniel-css-refactoring?expand=1#diff-c753a1396e7b4b55330c513675b157cc4aed6cb4508b6f00831bf44d20183c4d) as well as all the shadow styling that happens in custom web components themselves. The idea was to set as much of the general styling in the [core.css](https://github.com/UnitedLexCorp/SimpleTalk/compare/daniel-css-refactoring?expand=1#diff-c753a1396e7b4b55330c513675b157cc4aed6cb4508b6f00831bf44d20183c4d) file,  insisting that the shadow element sub-DOM would respond to this top level setting (to would allow for more explicitness and work well with controlling styles programmatically). This was largely successful, outside of `st-drawing` which interfaces with the internal canvas for sizing. In addition, I wanted the css to reflect our hierarchy structure, i.e. that `st-world` and `st-stack` are invisible, `st-card`'s take up the entire view-port of their parent (that being global view or a window) etc. 

Finally, the entire styling framework was refactored with the introduction of the [StyleProperty](https://github.com/UnitedLexCorp/SimpleTalk/compare/daniel-css-refactoring?expand=1#diff-c2dd5e63907f075364f2f774ef3baf2eeb51ce74c56cf051254b75bbacf3b90aR123) and a [styler](https://github.com/UnitedLexCorp/SimpleTalk/compare/daniel-css-refactoring?expand=1#diff-a7207430ea89f77a30ee56b19e948f99e4b298a5dba4a0f727ae30fbe0012c59) with its various utilities. The basic idea was to create a fully scriptable styling interface and also to completely decouple the SimpleTalk styling semantics with anything outside. For example, the [styler](https://github.com/UnitedLexCorp/SimpleTalk/compare/daniel-css-refactoring?expand=1#diff-a7207430ea89f77a30ee56b19e948f99e4b298a5dba4a0f727ae30fbe0012c59) function seen here is a specific `cssStyler` but in practice `StyleProperty` can take any function which transforms incoming SimpleTalk to something that a rendering engine can understand. We can have multiple system-wide stylers to play with, or use multiple within the same system. 

#### Notes ####
I've removed (most) of the style related props in PartView, replaced what was needed with the new `StyleProperty`'s in both PartView and its subclasses. At the moment I just stuck with keeping things to a minimum but also made decisions which seemed sensible. For example: removing all position and size from `PartView` since only some views really need this; adding these to views like window, field, etc... 

When a view is first instantiated, it can call a `setupStyleProperties()` from its `PartView` subclass. Note, this is not called automatically since not all views have styling properties and not all views might want to call these at instantiation (debatable). Instead of adding some sort of "type" attribute to all properties and using that, this method simply goes through all and picks out those that are instances of the `StyleProperty` class. This is fine for the moment I think but if we have more props types flying around, better design might be needed. 

With that said `PartView` has a `cssStyle` `BasicProperty` which is a dict/object containing css/JS valid key-values. This is what `StyleProperty` acts on when it is changed and it does so via the supplied styler. So in a sense `StyleProperty` is a particular type of `DynamicProperty`.  It is important to note that the `styler` alters the **entire** style object supplied to it, potentially altering multiple key-values, removing some, adding others, and does so in a way completely independent of the system or view or anything like that. So we are not married to a 1-1 mapping of ST style to css style.

This `cssStyle` property is what every view subscribes to, by simply updating `this.style` DOM element attribute. 

Halo and window moves now work with these properties not the DOM element style directly. This way there is no potential of mismatch (again with the exception of size for drawing, canvas..) 

I added a [`rotate` property](https://github.com/UnitedLexCorp/SimpleTalk/compare/daniel-css-refactoring?expand=1#diff-a7207430ea89f77a30ee56b19e948f99e4b298a5dba4a0f727ae30fbe0012c59R49) which translates to a css `transform: rotate(N)` as an example of what we can do, and also because it's fun. You can see that if we wanted to add say a "translate" ST property, our styler would need to update the `transform` key-value in `cssStyle` instead of just overwriting it, i.e. a rotate and translate would result in `transform: rotate(N) translate(X)` css. Anyhow we have a lot of room to play with. 

#### Tests ####
Tests have been written for the styler and related property change dynamics. Older tests have been updated/fixed to reflect the new structure. 

This is rebased over `master` as of #105 PR.

### DIFF
--- a/css/core.css
+++ b/css/core.css
@@ -24,7 +24,7 @@
     --drawing-default-width: 500px;
     --drawing-default-height: 300px;
     --field-default-width: 40vw;
-    --field-default-height: 20vh;
+    --field-default-height: 30vh;
 }
 
 /* Keyframe Animations */
@@ -39,8 +39,7 @@
 }
 
 .active {
-    background-color: black;
-    color: white;
+    outline: solid var(--palette-orange) 1px;
 }
 
 /* =World Styles
@@ -97,6 +96,10 @@ st-button:not(.editing):hover {
     cursor: pointer;
 }
 
+st-button:(.editing){
+    background-color: white;
+    color: black;
+}
 /* =Window Styles
  *----------------------------------------------------*/
 st-window {
@@ -145,6 +148,7 @@ st-svg {
  *----------------------------------------------------*/
 st-drawing {
     background-color: #1fe0; /* transparent */
+    position: absolute;
 }
 
 /* =Field Styles
@@ -152,6 +156,7 @@ st-drawing {
 st-field {
     width: var(--field-default-width);
     height: var(--field-default-height);
+    position: absolute;
 }
 
 
--- a/css/core.css
+++ b/css/core.css
@@ -19,6 +19,12 @@
     --palette-orange: #ff7c5d;
     --palette-red: #c70151;
     --palette-cornsik: #FFF8DC;
+    --container-default-width: 20vw;
+    --container-default-height: 15vh;
+    --drawing-default-width: 500px;
+    --drawing-default-height: 300px;
+    --field-default-width: 30vw;
+    --field-default-height: 20vh;
 }
 
 /* Keyframe Animations */
@@ -29,8 +35,7 @@
 
 /* Prerequisites */
 * {
-    box-sizing: border-box;
-    flex-grow: 1;
+    visibility: initial;
 }
 
 .active {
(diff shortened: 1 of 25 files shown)


darth-cheney (commented):
This all looks good to me. I don't see any issues when playing around.
  
One thing I realized while playing is that we don't really have a defined way to read a property value for the current part in context. For example, if I wanted to make an image part increase its top property by 10 each time I click it, how could I achieve that?
  
One way to do it might be to put `"the" <propNameLiteral> "of" <objectSpecifier?` as a grammar rule, which will attempt to lookup the named part property on the object. We should open an issue about it.
  
On a related note, I think we should have propNames in the `set` command not be string literals, but just prop name literals (no quotes). We that way we can also optionally pass the `"the"` when setting and it will be consistent (ie, `set the backgroundColor to "green"`). Or we could interpret both string literals and any sourceString, since the set command is so specific and well defined.

PR #110, opened 2021-02-12
Adding built in 'tell' command

darth-cheney:
## What
The tell sends a command message to some other specified part. This is
a way of sending messages outside of the message delegation chain.

It takes the form of:
   `tell <ObjectSpecifier> to <Command>`

For example:
```simpletalk
    tell first image of this card to loadFromUrl "http://some-image-source"
```
Note that the message is a deferred one. The tell command's first
argument is the id of the part that should execute it, and the second
argument is a correctly formatted command message object. The target,
once found, will simply send that second argument to itself. This
means that the part invoking the 'tell' command originally *will not
appear in the message's senders list*. This is what we'd expect, since
it is essentially "spoofing" the message send.

### DIFF
--- a/js/objects/System.js
+++ b/js/objects/System.js
@@ -1497,6 +1497,14 @@ System._commandHandlers['saveHTML'] = function(senders){
     anchor.parentElement.removeChild(anchor);
 };
 
+System._commandHandlers['tell'] = (senders, targetId, deferredMessage) => {
+    let targetPart = System.partsById[targetId];
+    if(!targetPart){
+        throw new Error(`Attempted to tell part id ${targetId}: no such part!`);
+    }
+    targetPart.sendMessage(deferredMessage, targetPart);
+};
+
 System._commandHandlers['startVideo'] = () => {
     if (video.srcObject !== null) {
         return;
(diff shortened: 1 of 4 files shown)


dkrasner:
@darth-cheney This is all simple, good and can be merged whenever. 

A few things: I am on the fence between "spoofing" and actually sending the message there. On the one hand, yes this is just getting around the delegation chain, on the other hand why should it not simply send a message/command along, i.e. why prevent parts/views from sending each other messages in this way? seems like the latter would allow to keep with the same framework we have and allow for better debugging. How would you debug a message/command if you have no idea where it came from

A few tangential things:
looping and variable tests are broken here and in `master` 

I can't get object selector by id to work, as in the screenshot below, either with systemObject or just "part"

image

I can do so by name, that works fine. 

I also can't identify a button that is not a direct descendant of a card. Maybe we want this to be the case though. For example, toolbox buttons live in stack.current > window > card. Do we want to be able to allow more "intelligent" selectors, i.e. those which will look for descendants.

Do we want to be able to shortcut things like this:
```
on click
   tell button "Add Button to Card" of first card of first stack of first window of current stack to click
end click
```

We would need some intelligent rules about it. For example, the look up by name would be greedy, as opposed to by id etc

btw just the fact that the above script works is totally awesome.

darth-cheney:
> Do we want to be able to shortcut things like this:
  
I think we should start thinking about using composition for cases like the one you've brought up. Here's what I mean: in your example you are finding a button in a window that requires a deeply nested specifier. But you could also just attach a handler in the script for the window that does the heavy lifting:
  
```simpletalk
-- In the script for the Window
on clickAddButton
    tell button "Add Button to Card" of first card to click
end clickAddButton

-- In the script of the button that is
-- not in the window
on click
    tell first window of current stack to clickAddButton
end click
```
  
We can go even further with this in thinking about the delegation chain and design of utility methods. For example, maybe in a Toolbox window the individual buttons themselves should delegate the things they are supposed to do to the Window's script. That way other parts can simply "tell" the window to perform those same activities. Taking just the add button to card example (not best example since adding parts is so easy/primitive):
```simpletalk
-- In the Toolbox window's script
on addButtonToCard
    add button to current card
end addButtonToCard

-- In the add button button inside the window
on click
    addButtonToCard -- will be delegated to window
end click

-- In any other button external to window
on click
    tell window "Toolbox" to addButtonToCard
end click
```

dkrasner:
Yeah I think that's a good patterns. Instead of composing the specifiers, just compose on the location of the scripts. 

It also reinforces the delegation chain and who does what where logic.

Issue #109, opened 2021-02-12
Execution Stack and Contexts


darth-cheney:
### The Problem ###
Our current ExecutionContext objects are not able to preserve their state when nested calls are made. This means variables are being overwritten. I attempted a hacky fix to this in #107 but as @dkrasner pointed out this will fail if you call two or more nested handler functions.
  
### Proposed Solution ###
We should add an ExecutionStack object to the System, and have each handler push / pop configured ActivationContext objects to it. That way we will always know what is "current". This might also help us compose a primitive debugger. I will take this up next since it is critical to a demo release.

darth-cheney:
This has been resolved by #111. Closing.

PR #111, opened 2021-02-15
Execution Stack and Activation Context

darth-cheney:
## What ##
This PR implements a replacement of our current `ExecutionContext` structure for dealing with local variables when executing message handlers.
  
## The Problem ##
A quick summary of the problem can be seen in the following example button script:
```simpletalk
on click
    put 1 into myVar
    doSomethingElse
    answer myVar
end click

on doSomethingElse
   put 2 into myVar
   doYetAnotherThing
end doSomethingElse

on doYetAnotherThing
    put 3 into myVar
end doYetAnotherThing
```
What we would expect to happen in the above script is to `answer 1`. But with the current setup, in fact, we get `answer 3`. This is because the current execution context implementation cannot account for calls that are n-deep. Additionally, these contexts were being attached to the part instances themselves and then being overwritten each time some nested calls were made.
  
### Overview ###
The new implementation provides the `System` object with its own execution stack, an instance of [`ExecutionStack`](https://github.com/UnitedLexCorp/SimpleTalk/blob/eric-stack/js/objects/ExecutionStack.js#L11). The key thing to know is that `System.executionStack.current` will always refer to the top of the stack.
  
We also provide a new type of object called [`ActivationContext`](https://github.com/UnitedLexCorp/SimpleTalk/blob/eric-stack/js/objects/ExecutionStack.js#L57). This is something akin to what other environments might just call an "activation," what Smalltalk talks a Context or MethodContext, and can also be thought of as a kind of stack frame. This object holds a dict of variable names to values for the current handler invocation, as well as references to things like the part on which it is executing, the message object that invoked it, and even the concrete JS function that is the handler.
  
The idea is simple enough: whenever a new message handler is invoked, an `ActiveContext` object should be initialized for it and pushed to the top of the execution stack. When the handler has finished, it should pop the top of the stack off.
  
The advantage here is that the various semantic operations in the interpreter need to know as little as possible in order to ascertain the correct values, especially variable mappings. For example, here is how the interpreter looks up a variable now:
```javascript
        variableName: function(letterPlus, optionalDigits){
            // Lookup the variable in the part's
            // current execution context
            // If the variable is not a key on the object,
            // we throw an error: this means the variable has not yet
            // been defined but is being looked up.
            let value = systemContext.executionStack.current.get(this.sourceString);
            if(value == undefined){
                throw new Error(`Variable ${this.sourceString} has not been defined`);
            }
            return value;
        },
```
This particular operation -- invoked by `variableName.interpret()` -- only needs to know about accessing the current stack top in order to get the correct variable bindings.
  
Because this is all just a stack, we can now deal with n-deep nested or recursive calls while maintaining the integrity of local variables.
  
### Question: Where? ###
This brings up an important question: where do the ActivationContexts actually get initialized, and where is the pushing/popping taking place?
  
Initially, I had this all happening inside the interpreter itself whenever a new handler function was being invoked. This had two main drawbacks:
1. We would not be pushing built-in/hard-coded command handlers to the stack because they are not actually interpreted (they would be invisible to the stack);
2. We cannot add some useful information to ActivationContexts, such as the message that invoked the handler or a reference to the handler function itself
  
I decided, then, to switch gears and make this all happen in two places: the `receiveCommand` methods of both System.js ([see here](https://github.com/UnitedLexCorp/SimpleTalk/blob/eric-stack/js/objects/System.js#L361)) and Part.js ([see here](https://github.com/UnitedLexCorp/SimpleTalk/blob/eric-stack/js/objects/parts/Part.js#L385)). For now this seems to work well.
  
### Important To Know: `putInto` ###
We have one primitive / built-in command handler at the System level for variable assignment: `putInto`. Because it is implemented as a command, it pushes itself to the top of the stack when it is invoked. However, the whole point of this particular primitive is to set a variable in some other context that is not itself. To accomplish this, we provide the `ExecutionStack` object with a `previous` getter that returns the second to the top item on the stack (if present). You can see [here](https://github.com/UnitedLexCorp/SimpleTalk/blob/eric-stack/js/objects/System.js#L930) how our implementation makes use of it. This allows the handler to set a variable in the context from which it was invoked.
  
Something to consider here is the possibility of removing the `putInto` command handler from System entirely. If we want, we can have the interpreter itself simply make sure the values are assigned to `executionStack.previous` in the operation that interprets `putInto`, rather than sending any kind of message. This is something we should discuss.
  
## Visual Example ##
The following sketch shows the variable bindings and stack configurations on ActiveContexts during the invocation of `click` in the following example script:
```simpletalk
on click
    put 1 into myVar
    doSomethingElse
    answer myVar
end click

on doSomethingElse
    put 2 into myVar
end doSomethingElse
```
ExecutionStackExample-1
ExecutionStackExample-2
  
## Issues ##
There is an issue with the repeat looping tests, which unfortunately require being able to look up variable values after things have executed. This is no longer possible since the stack pops those contexts and forgets about them. I have marked these tests to skip for the moment, but we should think about how to reimplement them. Using global variables instead is the most obvious direction to go.
  
## Future Work ##
If you squint, this work seems like it might be the beginning of a message handler debugger. I think it leaves us with a solid foundation for experimenting with such a thing.
  
We will also want to be aware of possible performance implications of this new pattern. If we are doing intensive activities like animations by actually setting and getting properties, the overhead might be too large for smooth action. It's something to keep an eye out for.
  
## Closes ##
#109


ApproximateIdentity:
So basically you are introducing a scope local to a handler definition. What are the formal scoping rules we are working under? This "fixes" the issue if our rule is something like "variables defined in each handler are local to this handler", but the old version was correct if "variables are global". Do we have the semantics written up somewhere? Or is this basically following hypercard like semantics and there they have something similar going on?

darth-cheney:
@ApproximateIdentity Your assumption is correct -- variables defined in each handler are local to that handler. Same goes for argument variables. However, we also provide a way to set global variables (see [here](https://github.com/UnitedLexCorp/SimpleTalk/blob/master/js/ohm/simpletalk.ohm#L207)). These scoping rules are identical to HyperTalk/Hypercard

js/objects/ExecutionStack.js
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    get(varName){
+        let localValue = this.getLocal(varName);
+        if(localValue !== undefined){
+            return localValue;
+        };
+        // otherwise try to return a global
+        // variable
+        return window.System.executionStack.getGlobal(varName);
```
ApproximateIdentity, line 78: It looks like this will return `undefined` if it isn't set. Is that what we want? No error?

darth-cheney: Yeah. The error is handled when the variable is looked up by the interpreter, [here](https://github.com/UnitedLexCorp/SimpleTalk/blob/master/js/ohm/interpreter-semantics.js#L861).

ApproximateIdentity: Hmm sorry I didn't fire up this branch and was only looking at the code. It looked like it just called the `ExecutionStack.getGlobal()` method in this file:

```
    getGlobal(varName){
        return this._globals[varName];
    }
```

which is why I thought it would just return undefined.

darth-cheney: You are right though, at that point in the code it would return undefined. Perhaps we should throw these errors from here rather than in the interpreter (we would do the same for local var lookups in an ActiveContext). You think that's a better way to go?
  
Also I had provided a link at the wrong branch in my previous response. [This](https://github.com/UnitedLexCorp/SimpleTalk/blob/eric-stack/js/ohm/interpreter-semantics.js#L855) is the correct spot.

ApproximateIdentity: Ah I see that's why I was confused. Honestly I don't think it matters really as long as you're consistent. Handling it somewhere else is fine too. You don't need to move things around because I'm unfamiliar with the code.

dkrasner: Yeah I think at this point we should send/throw a `ReferenceNotKnown` message which can then be handled in our error message function, then marked up in the script editor (or later debugger)

dkrasner:
> There is an issue with the repeat looping tests, which unfortunately require being able to look up variable values after things have executed. This is no longer possible since the stack pops those contexts and forgets about them. I have marked these tests to skip for the moment, but we should think about how to reimplement them. Using global variables instead is the most obvious direction to go.

What about introducing part-context variables, as suggested in [this thread](https://kfitter.slack.com/archives/C016RQPJX6X/p1613070630001400) These could be subsequently shared amongst command handlers in any single part. 

Otherwise I am not sure how we could make even something like a click counter without global vars.

dkrasner:
you also fixed the bug talked about in  #107  where global wasn't somehow overwritten

```
on click
    put 0 into global counter
    answer counter
    put (1 + counter) into counter
    answer counter
end click
```

this does what you expect, i.e. always sets "counter" to 0

But we do want to be able to see if variables have been set. For example,

```
on click
   if not counter then put 0 into global counter
   answer counter
   put (counter + 1) into global counter
end click
```

at the moment throws a `Factor_notFactor` `missingSemanticAction` ohm error. 

you could get around this by creating the counter set to 0 in another button context or in a previous run but that's awkward

for example:

```
-- run 1
on click
    put 0 into global counter
end click
```

```
-- run2
on click
   answer counter
   put (counter + 1) into global counter
end click
```

dkrasner:
> Something to consider here is the possibility of removing the `putInto` command handler from System entirely. If we want, we can have the interpreter itself simply make sure the values are assigned to `executionStack.previous` in the operation that interprets `putInto`, rather than sending any kind of message. This is something we should discuss.
> ##

My take is that since `put .. into` is a core ST language command it should result in a message send like everything else. At least for debugging purposes and framework coherence

dkrasner:
> ### Question: Where?
> 
> This brings up an important question: where do the ActivationContexts actually get initialized, and where is the pushing/popping taking place?
> 
> Initially, I had this all happening inside the interpreter itself whenever a new handler function was being invoked. This had two main drawbacks:
> 
>     1. We would not be pushing built-in/hard-coded command handlers to the stack because they are not actually interpreted (they would be invisible to the stack);
> 
>     2. We cannot add some useful information to ActivationContexts, such as the message that invoked the handler or a reference to the handler function itself
> 
> 
> I decided, then, to switch gears and make this all happen in two places: the `receiveCommand` methods of both System.js ([see here](https://github.com/UnitedLexCorp/SimpleTalk/blob/eric-stack/js/objects/System.js#L361)) and Part.js ([see here](https://github.com/UnitedLexCorp/SimpleTalk/blob/eric-stack/js/objects/parts/Part.js#L385)). For now this seems to work well.

I think this is right. Let the part (or System) deal with its own context for the moment. If this becomes too decentralized we can adjust as needed. The current implementation is simple enough to adjust as needed

js/objects/ExecutionStack.js
```diff
@@ -0,0 +1,93 @@
...
+    get previous(){
+        if(!this._stack.length >= 2){
+            return null;
+        }
+        return this._stack[this._stack.length - 2];
+    }
+};
+
+class ActivationContext {
+    constructor(messageName, part, incomingMessage, handlerFunction){
```
dkrasner, line 58: this is great. We can get all the debugging info we need from one place, and if need be we can enrich the ActivationContext

dkrasner:
> ## Future Work
> 
> If you squint, this work seems like it might be the beginning of a message handler debugger. I think it leaves us with a solid foundation for experimenting with such a thing.
> 

Yeah definitely. I think it will also set a cleaner framework for remote and collaborative scripting/debugging etc. 

> We will also want to be aware of possible performance implications of this new pattern. If we are doing intensive activities like animations by actually setting and getting properties, the overhead might be too large for smooth action. It's something to keep an eye out for.

agreed, and the same is true of our roundabout way of dealing with web events but we'll deal with such things when they become issues

js/objects/ExecutionStack.js
darth-cheney: I will add some official error types for variable lookup and push them to this branch

darth-cheney:
> What about introducing part-context variables, as suggested in [this thread](https://kfitter.slack.com/archives/C016RQPJX6X/p1613070630001400) These could be subsequently shared amongst command handlers in any single part.
> 
> Otherwise I am not sure how we could make even something like a click counter without global vars.
  
Let's kick the can on this and discuss it further. If we did something like arbitrary associations added to parts, we would want to make sure that the language has the appropriate data structures implemented first (which it doesn't). Note also that we can start to use "composition" of the relationship of current parts. For example, an invisible field on a card can be used to store a and read a value. This is a method authors would have used in Hypercard.
  
> at the moment throws a Factor_notFactor missingSemanticAction ohm error.
  
I'll look into this. There are several places in the grammar now that are limited to just accepting literals or variable names when, in fact, we want them to accept any Factor (anything that when evaluated will result in a value).

dkrasner:
> > What about introducing part-context variables, as suggested in [this thread](https://kfitter.slack.com/archives/C016RQPJX6X/p1613070630001400) These could be subsequently shared amongst command handlers in any single part.
> > Otherwise I am not sure how we could make even something like a click counter without global vars.
> 
> Let's kick the can on this and discuss it further. If we did something like arbitrary associations added to parts, we would want to make sure that the language has the appropriate data structures implemented first (which it doesn't). Note also that we can start to use "composition" of the relationship of current parts. For example, an invisible field on a card can be used to store a and read a value. This is a method authors would have used in Hypercard.
> 
> > at the moment throws a Factor_notFactor missingSemanticAction ohm error.
> 
> I'll look into this. There are several places in the grammar now that are limited to just accepting literals or variable names when, in fact, we want them to accept any Factor (anything that when evaluated will result in a value).

for the moment I simply meant to have "global"  <- system global, "local" <- scoped for part, and default which would be handler scoped 

the rest of the discussion in that thread I think we can kick down the road until things on the interpreter/stack side stabilize a bit more...

darth-cheney:
> you could get around this by creating the counter set to 0 in another button context or in a previous run but that's awkward
  
I just came across [this video](https://youtu.be/fTPcA6w3Hrs?t=540) of a kid describing the exact same counter problem in Hypercard. Seems he used the Field method.

PR #113, opened 2021-02-18
transparent

dkrasner:
### Main Points ###

A simple setup for "transparent" and "visible" part properties. The properties and styler are fine and trivial to set up. 

I added these to the color-wheel (there is an [option](https://github.com/UnitedLexCorp/SimpleTalk/compare/daniel-transparent?expand=1#diff-2433c8a1dcdb9e45dfd097ee0a4178a4d44f6a2f3f5f121f6f61b97644b962e9R166) to not include them). Since the color-wheel was set up to send custom events on color select, b/c it is not aware of its target, I decided to stick to the same framework. I don't think it's very good though and forces a lot more code than necessary. For example you need to write event handlers for each part-view that has a color wheel acting on it. 

I think better would be to create editors for each part that are fully aware of their target, like button editors, and have whatever widgets they use also be aware of the target and then simply send a message to the corresponding part property. 
Similarly drawing attributes like color, brush stroke etc should be part props on the drawing part-model and set accordingly. But that's for another PR. 

One thing to notice is that toggling the transparent checkbox can technically make the color-wheel also transparent. This is b/c the underlying cssStyler converts this property (if set to true) to `visibility:hidden` which is inherited by the color-wheel if it is setup as a DOM descendant of the target part-view. I got around this by forcing `visibility: visible` on the color-wheel: host. But the same can't be said for the "visible" property which at the moment sets `display: none` and that you can't get around. Of course you could use some other css mapping but that I don't think is the point. Probably best to remove editors from their respective views all-together.

### DIFF
--- a/css/core.css
+++ b/css/core.css
@@ -157,6 +157,7 @@ st-field {
     width: var(--field-default-width);
     height: var(--field-default-height);
     position: absolute;
+    background-color: var(--palette-cornsik);
 }
 
 
--- a/js/objects/parts/Button.js
+++ b/js/objects/parts/Button.js
@@ -54,11 +54,15 @@ class Button extends Part {
         );
         this.partProperties.newStyleProp(
             'background-color',
-            null,
+            "rgba(255, 234, 149, 1)", // var(--palette-yellow)
+        );
+        this.partProperties.newStyleProp(
+            'background-transparency',
+            1,
         );
         this.partProperties.newStyleProp(
             'text-color',
-            "rgba(0, 0, 0, 1)",
+            "rgba(0, 0, 0, 1)", // black
         );
         this.partProperties.newStyleProp(
             'text-style',
--- a/js/objects/parts/Button.js
+++ b/js/objects/parts/Button.js
@@ -65,7 +65,7 @@ class Button extends Part {
             'plain',
         );
         this.partProperties.newStyleProp(
-            'text-tranparency',
+            'text-transparency',
             1,
         );
         this.partProperties.newStyleProp(
--- a/js/objects/parts/Button.js
+++ b/js/objects/parts/Button.js
@@ -95,6 +95,10 @@ class Button extends Part {
             'rotate',
             null,
         );
+        this.partProperties.newStyleProp(
+            'transparent',
+            false,
+        );
         // TODO tbd props
         this.partProperties.newBasicProp(
             'autoHilite',
--- a/js/objects/parts/Button.js
+++ b/js/objects/parts/Button.js
@@ -58,19 +58,19 @@ class Button extends Part {
         );
         this.partProperties.newStyleProp(
             'text-color',
-            null,
+            "rgba(0, 0, 0, 1)",
         );
         this.partProperties.newStyleProp(
             'text-style',
             'plain',
         );
         this.partProperties.newStyleProp(
-            'name-visible',
-            true,
+            'text-tranparency',
+            1,
         );
         this.partProperties.newStyleProp(
-            'visible',
-            true,
+            'hide',
+            false,
         );
         // setting width and height to null
         // effectively forces to the default size
@@ -96,8 +96,8 @@ class Button extends Part {
             null,
         );
         this.partProperties.newStyleProp(
-            'transparent',
-            false,
+            'transparency',
+            1.0,
         );
         // TODO tbd props
         this.partProperties.newBasicProp(
--- a/js/objects/parts/Container.js
+++ b/js/objects/parts/Container.js
@@ -48,6 +48,14 @@ class Container extends Part {
             'rotate',
             null
         );
+        this.partProperties.newStyleProp(
+            'visible',
+            true,
+        );
+        this.partProperties.newStyleProp(
+            'transparent',
+            false,
+        );
         this.setupStyleProperties();
     }
 
--- a/js/objects/parts/Container.js
+++ b/js/objects/parts/Container.js
@@ -49,12 +49,12 @@ class Container extends Part {
             null
         );
         this.partProperties.newStyleProp(
-            'visible',
-            true,
+            'hide',
+            false,
         );
         this.partProperties.newStyleProp(
-            'transparent',
-            false,
+            'transparency',
+            1.0,
         );
         this.setupStyleProperties();
     }
--- a/js/objects/parts/Container.js
+++ b/js/objects/parts/Container.js
@@ -29,6 +29,10 @@ class Container extends Part {
             null,
         );
         this.partProperties.newStyleProp(
+            'background-transparency',
+            1,
+        );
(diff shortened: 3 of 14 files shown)


darth-cheney:
> I don't think it's very good though and forces a lot more code than necessary. For example you need to write event handlers for each part-view that has a color wheel acting on it.
  
I was probably thinking this would be a "general purpose" webcomponent when I first made it, but we don't need that kind of thing. We can definitely make this simpler. We can probably even implement a lot of it in ST down the road.
  
> I think better would be to create editors for each part that are fully aware of their target, like button editors, and have whatever widgets they use also be aware of the target and then simply send a message to the corresponding part property.
  
>  Probably best to remove editors from their respective views all-together.
  
Yes. Definitely. I believe Windows have a part property that specifies their "target" part, which can then be accessed by other parts inside. Maybe we can do something similar.
  
> A simple setup for "transparent" and "visible" part properties. The properties and styler are fine and trivial to set up.

Biggest issue I have here is the distinction between ST's notion of "visibility" and "transparent" vs what the CSS can do. In CSS land, there are three ways to make something "invisible":
  
1. Set `visibility: hidden`. This has the important distinction that the element still takes up space in any layout, it's just invisible. It will affect any layouts that are not absolutely positioned. That is how it is (crucially) different from `display: none`.
2. Set `display: none`. The element is not computed in any layout and therefore not shown in the display at all. It's as if it isn't there.
3. `opacity: 0` -- This sets the element's opacity. It can take a value between `0` and `1`. Setting the opacity to zero will produce similar results to `visibility: hidden`, but there is an extra advantage in that we can specify some opacity.
  
I bring this up to suggest that we get rid of 1 and 2 altogether, and have our ST properties be a mapping to opacity. In this case, what we call "transparency" is just `1 - opacity`. This might even work in JS of we continue to set transparent to `true` or `false` because these will be interpreted as 1 and 0 in any math operation.
  
The upshot is that the color picker could now have a slider for transparency that produces values between 0 and 1. Or any halo can just have a button that only handles transparency. Or both -- whatever.
  
This would also be good for things like animation, where a "fade-in" effect is just a step message that gradually decrements the transparency property. Etc, etc.
  
What do you think?

dkrasner:
yeah agreed, I think transparent should not be bool and become transparency as you say mapping to css: opacity. And i like the slider - sliders are always good. 

the reason I added visible: bool, was due to our discussion of variables, named values, stores for them and such. If you want a field to solely store some value, like a button click counter for example, you don't want it to take room in a layout. Now if the layout is absolute then it doesn't matter but if it is flex/grid then you want this field to really disappear. So I thought add this in. Doesn't mean it needs to be on the color picker. What do you think?

As far as building out editors. The current buttonEditor has a target (note it's just a view, has no model), but I really wanted it to be a window with stuff in it, or a container. But ST was not ready for it and I wanted a quick editor that did things to the button. From a "what is a part" perspective these editors should probably be containers, since a window is a 'view onto something' but that is highly debatable. Regardless I think we should pick one and, once we have flex layout, rewrite the current button editor and the rest of them. 

We also need some widgets which might or might not be parts: colorwheel, sliders, checkbox. On the one hand they really could just be views, like Halo is; on the other hand if we want to script them in then our current setup asks for the part/model-view (`add button to card` requires that the button is a part), although this can be changed...

darth-cheney:
>  If you want a field to solely store some value, like a button click counter for example, you don't want it to take room in a layout. Now if the layout is absolute then it doesn't matter but if it is flex/grid then you want this field to really disappear. So I thought add this in. Doesn't mean it needs to be on the color picker. What do you think?
  
Yeah we definitely need to have this too. I'm just worried about the user-level distinction between "transparent" and "invisible", since a transparency of 1 technically makes something invisible. Maybe we should call the non-layout option "hidden" or "hide" and it takes a boolean?
  
> Regardless I think we should pick one and, once we have flex layout, rewrite the current button editor and the rest of them.
  
Yeah for sure.

dkrasner:
"hide"   good

dkrasner:
i'll make these changes and add the slider to replace those check-boxes   then merge this in. Don't want to spend to omuch time on the editing parts since i'd like to redo all of it

darth-cheney:
Sounds good

dkrasner:
@darth-cheney take a look at the last 2 commits above and maybe play around a bit. 

I think using transparency for both text and background (part-view) is pretty natural.

dkrasner:
@darth-cheney take a look at the last 3 commits above and maybe play around a bit. 

I think using transparency for both text and background (part-view) is pretty natural. 

I also added a callback for drawing. Some of the drawing editing is a bit buggy and also doesn't fit into the our styleProps setup (much like the Field #117) but i think we can kick this down the road to a reworking of all the editors

darth-cheney:
A few things I noticed playing with the recent changes:
1. The color picker for Drawing cannot be closed
2. Attempting to change the background color of the Field actually changes the transparency of the whole element, so the whole thing (text included) is fading rather than just the background:
Screenshot from 2021-02-23 08-27-51
3. There are some style-related errors in the console when switching a field from SimpleTalk mode to Bravo mode

dkrasner:
> A few things I noticed playing with the recent changes:
> 
>     1. The color picker for Drawing cannot be closed
> 
been the case for a while. You could hack it in which i did, but i think drawing needs a new editor setup. I dropped fixing it...
>     2. Attempting to change the background color of the Field actually changes the transparency of the whole element, so the whole thing (text included) is fading rather than just the background:
>        Screenshot from 2021-02-23 08-27-51

as intended. This is how you make an element transparent - no? 
> 
>     3. There are some style-related errors in the console when switching a field from SimpleTalk mode to Bravo mode
yeah this is related to #117  - the whole thing, like drawing, needs to be refactored to be compatible with styling props

darth-cheney:
> This is how you make an element transparent - no?
  
Yeah that's how you make an element transparent, but I thought the purpose of that particular button in the Field was to make just the background transparent (and leave the foreground text as-is). Right now there's no way to do that.

dkrasner:
> > This is how you make an element transparent - no?
> 
> Yeah that's how you make an element transparent, but I thought the purpose of that particular button in the Field was to make just the background transparent (and leave the foreground text as-is). Right now there's no way to do that.

yeah i guess we need some other place to make an element transparent. Lets kick it down the road, when editors come into play. I am not sure where I'd put this slider now for the entire element and there would need to be some refactoring of the background color style (similar to how text-color) is done in order to make it affect the background transparency...

thoughts?

dkrasner:
image
image


@darth-cheney ok this is not bad 

note you can still set the `transparency` prop using any command, there is just not a visual slider element in the editors for it. Also now everything color related is rgba (with a small colorName -> rgba dict for convenience). Personally I like this, keeps things uniform.

PR #116, opened 2021-02-22
New Serialization

darth-cheney:
## What ##
This PR represents a reimplementation of SimpleTalk World serialization/deserialization. It allows us to both save a full HTML file with the state of our Parts fully preserved, and to reload such state from a simple, flat JSON object that is itself encoded in the HTML file.
  
In the process of refactoring serialization, we also made some other changes worth noting upfront:
1. Toolbox is no longer a special object -- it is simply serialized into the new bootstrap file like any other SimpleTalk Part.
2. The `add <partName> to <object>` scripted command can now take an ObjectSpecifier as its target. This was overlooked before.
3. Various smaller changes, some of which will be discussed below.
  
## Implementation ##
A full serialization of the world is a JSON string with the following keys:
| key | description |
|------|-----------------|
| **`parts`**| A dict of Part ids to a serialization of those parts (see below) |
|**`currentCardId`**| The ID of the current card at the time of serialization |
|**`currentStackId`**| The ID of the current stack at the time of serialization |
  
An individual "part serialization" (stored in `parts` above) is simply a JSON representation of a given Part's PartProperties, along with a flat representation of its subparts. By "flat" we mean the subparts will simply be serialized as an array of id values and not nested serializations of the subparts themselves.
  
The full serialization is stored in a `<script>` tag at the bottom of the bootstrap file. This element will have an id of `serialization`.
  
We run a serialization any time that a new model is created. We should discuss other times we think this should automatically happen, and perhaps whether or not we want to save snapshots or something like that in browser storage. TBD.
  
#### A Note on IDs ####
When deserializing, we do **not** guarantee preservation of IDs in the new models. Doing so has led to incredibly difficult clashes and it is easier and arguably better to just give out new random IDs. If we ever change our ID schema we might want to revisit this point.
  
### The New Toolbox ###
As mentioned at the top, we have replaced the special System object Toolbox with a simple serialization of a Window part that contains all Toolbox functionality.
  
#### How It Was Made ####
It's important -- and perhaps interesting -- to understand how this new Toolbox was made. *At no point was the browser console used nor was anything hard-coded in the System.* Creating the Toolbox was done entirely by scripting buttons with basic SimpleTalk scripts and building from there. The final step was to add the "Save" button to the Toolbox itself, then click it and save the result. This gave us (after several iterations) the new `bootstrap.html` file.
  
#### Closing Windows ####
I have updated the close button of window to send the `windowClose` message to itself. There is a built-in private command handler that simply deletes the window model from the System. However, users can override this by scripting a handler for `windowClose` in a given window's script. This is exactly what the new Toolbox does (see below), and it's how we should start writing the scripts for our Stacks, Cards, and even the WorldStack
  
#### Current Caveats ####
First, I have disabled shift-clicking to open the Toolbox. We didn't notice it before, but this event was not being captured correctly. We need to adjust it so that it doesn't fire on every shift click in the viewport (for example, halo activation clicks) which actually was happening before.
  
Second, you might notice that you cannot close the Toolbox without throwing an error. This is because of the new `windowClose` handler on the Toolbox, which simply sets its `hide` property to `false`. However, there is no `hide` property! Not yet at least -- this is forthcoming in @dkrasner 's PR
  
### Changes to `newModel` ###
One of the refactorings that took place during this cycle was to change the function signature of `System.newModel`. Previously it expected lots of extra arguments to determine the context in which a new Part should be created. This has not been superseded by the interpreter and ObjectSpecifiers.
  
Simplifying this command handler and changing its function signature required altering other parts of the code that were calling it directly in the old fashion. I have changed these. Additionally, any tests were also updated accordingly.
  
### About Tests ###
There were *lots* of stale tests that had certain expectations about the initial state of an empty initialized System -- including and especially the presence of a Toolbox. The new implementation means there is no Toolbox without the bootstrap file. This means the test assumptions are now wrong; I have gone in and adjusted or removed these assumptions where needed.
  
In other areas, like the tests of the Toolbox itself, I have chosen to skip the tests altogether.
  
## What You Should Do ##
Make stacks. Save them. Reload the saved versions. And keep doing this. Report any bugs you find here in this PR and we can roll from there. Enjoy!


ApproximateIdentity:
On thing that would be nice is if the json-encoded object were serialized in a more reader-friendly style with newlines and indentations. Then if you saved a file, clicked a button, and then saved it again, you should presumable be able to diff the two and see the difference in the generated json object right away. At least it's something I would like.

ApproximateIdentity:
```
When deserializing, we do not guarantee preservation of IDs in the new models.
Doing so has led to incredibly difficult clashes and it is easier and arguably better
to just give out new random IDs. If we ever change our ID schema we might want
to revisit this point.
```
Does that mean that we won't be able to preserve handlers that make explicit use of IDs? The hand detection model stuff uses them:

https://github.com/UnitedLexCorp/SimpleTalk/blob/master/js/objects/System.js#L1627-L1637

Of course the design of the hand detection model should probably come last in all considerations here so if that just needs to be changed, then that's fine, but I just wanted to make sure I understood the implications of the ids changing.

darth-cheney:
@ApproximateIdentity : 
> I explicitly pass a recipient ID to send the message back to in the current design. Does this break?
This *will* affect scripts that reference specific IDs. However, if it is not in a script you should be OK.
  
> On thing that would be nice is if the json-encoded object were serialized in a more reader-friendly style with newlines and indentations. Then if you saved a file, clicked a button, and then saved it again, you should presumable be able to diff the two and see the difference in the generated json object right away. At least it's something I would like.
  
I just pushed a commit that does this. You will have to re-save the bootstrap page to see the default in prettification, but subsequent serializations should be prettier.

ApproximateIdentity:
Yeah the scripts I wrote to pass hand data were referencing object specifically by their ids so I guess that will break. That's okay just something to keep in mind. I can do it differrently when integrating the current stuff. Don't think it matters at all for the "push hand in for click" way of doing things.

dkrasner:
I am getting the following error
image


i'll upload the snapshot to slack

js/ohm/interpreter-semantics.js
```diff
@@ -743,6 +735,21 @@ const createInterpreterSemantics = (partContext, systemContext) => {
...
+        /**
+         * An ObjectSpecifier without an annotated
+         * rule means it was interpreted as just
+         * a TerminalSpecifier of some sort.
+         * However, we need to extract the id
+         * and return that result, since that is what is
+         * expected of all interpreted ObjectSpecifiers
+         */
+        ObjectSpecifier_singleTerminal: function(terminalSpecifier){
+            console.log(terminalSpecifier.sourceString);
```
dkrasner, line 747: remove?

dkrasner:
@darth-cheney went through the code all makes sense...

looks like some properties are not being preserved:

original:

image


serialized: 
image

I would have expected that button to have transparent background, bigger size etc

darth-cheney:
@krasner -- Yes, those should be preserved. Drawings are also not being preserved (but Images seem to be OK). I'll look into it

js/ohm/interpreter-semantics.js
darth-cheney: Done

darth-cheney:
Ok @dkrasner I have fixed that issue, which was pretty dumb. I've been able to serialize some stuff here without issue.

Issue #117, opened 2021-02-22
Field and styleProps


dkrasner:
### Main Points ###

Field is not compatible with the new style props (outside of text and background as seen in this pr #113)

PR #119, opened 2021-02-23
doIt

dkrasner:
### Main points ###

Here we add the possibility of running command lines from any field view element. This could be a field element that contains a script on some target part, or it could be a standalone field. 

The basic idea is that anything you can write in, can become your "playground." You just type whatever statement line you want, select it and press the "doIt" context menu button that pops up. 

The way this works is that the selected command lines are wrapped in a "doIt" handler, compiled and a "doIt" command is sent to the underlying model. You can see that all play out in [these few lines](https://github.com/UnitedLexCorp/SimpleTalk/compare/daniel-doit?expand=1#diff-48ed5cd662d43c127d7e1ce0f6f561e5d6a805cb07e182d40b4f3ddea2dbb1acR559)

#### Known issues ####
* selected statement lines are run in the context of the field in which they are written. This means that if you select and run `add button to this card` it will add the button to card parent of the field you are selecting in which might or might be the parent of the owner of the script in that same field. I think this should be handled when we make proper editors and sort out how targets (of fields that have scripts) are attributed
* with the above said if you try to run style changing commands like, `set "background-color" to "black"`, they will fail because fields, and drawings, are not compatible with ST styling at the moment (see #117 for example)
* there is some weirdness interacting with basic textarea/contenteditable divs (such as copy, backspace) and the contextMenu which comes up automatically. Maybe I should do something else, like `if selected text && alt-LeftClick` then bring up the context menu...
* the context menu is a single button at the moment, but i assume it will get more involved
* you can't select entire handlers, only statement lines. The reason I chose not to go down this route is b/c handlers might have arguments defined so what does it mean to run a handler without passing in the required args? and if you want to a run a handler without the args, just select the statement lines it contains.
* if you are in a script field an there is an error in the statement line(s) the error will not be handled properly given the context issues above

### DIFF
--- a/js/objects/views/FieldView.js
+++ b/js/objects/views/FieldView.js
@@ -490,11 +490,12 @@ class FieldView extends PartView {
             // the context menu interaction can prevent backspace from
             // removing text as it intuitively should
             if(event.key==="Backspace"){
-                this.textarea.innerHTML = "";
+                document.getSelection().deleteFromDocument();
+                let innerHTML = this.textarea.innerHTML;
                 this.model.partProperties.setPropertyNamed(
                     this.model,
                     'htmlContent',
-                    ""
+                    innerHTML
                 );
             };
             this.closeContextMenu();
--- a/js/objects/views/FieldView.js
+++ b/js/objects/views/FieldView.js
@@ -220,7 +220,6 @@ class FieldView extends PartView {
         this.onInput = this.onInput.bind(this);
         this.onClick = this.onClick.bind(this);
         this.onKeydown = this.onKeydown.bind(this);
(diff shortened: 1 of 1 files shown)


dkrasner:
**update**: ok i got too annoyed at how it was working and instead of the automatic context menu popping up you now need to press `alt` then select whatever. 

This is much cleaner from a code perspective, just adding one clause in the already `onClick` handler and it doesn't break any of the browser natural textarea functionality. 

see [last commit](https://github.com/UnitedLexCorp/SimpleTalk/pull/119/commits/1d604039245f1e72c90936efdfa7521381995d81)

PR #120, opened 2021-02-24
Image from Local File

darth-cheney:
## What ##
This PR implements a built-in command that allows the Image part to load data from a local file on the user's filesystem.
  
## Implementation ##
It's rather simple: when the Image part receives the `loadImageFromFile` message, it will create an invisible file input element in the DOM, click it, and update the Image Part's `imageData` part property to be the data from the file.
  
Note also that the Image Part's `src` property and `mimeType` will also be updated. We assume that an Image Part whose `src` is null is a locally-loaded image. This is for future serialization purposes -- we want to make sure that when a page loads, we *don't* attempt to load from the set `src`, but rather simply display the cached `imageData` serialization.
  
## Testing ##
Because this involves user interaction, we really need to test it ourselves. Here are some steps to test:
1. Open the bootstrap page
2. Create a new image
3. Create a new button
4. Add the following to the button's script:
```simpletalk
on click
    tell first image of this card to loadImageFromFile
end click
```
5. Select an image file from your computer
  
The image should load into SimpleTalk without error. Try both binary images and SVGs to make sure it's working

### DIFF
--- a/js/objects/parts/Image.js
+++ b/js/objects/parts/Image.js
@@ -57,9 +57,11 @@ class Image extends Part {
 
         // Private command handlers
         this.setPrivateCommandHandler("loadImageFrom", this.loadImageFromSource);
+        this.setPrivateCommandHandler("loadImageFromFile", this.loadImageFromFile);
 
         // Bind component methods
         this.loadImageFromSource = this.loadImageFromSource.bind(this);
+        this.loadImageFromFile = this.loadImageFromFile.bind(this);
     }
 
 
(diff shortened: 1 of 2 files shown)


js/objects/parts/Image.js
```diff
@@ -108,9 +110,39 @@ class Image extends Part {
...
     }
 
+    loadImageFromFile(){
+        let filePicker = document.createElement('input');
+        filePicker.type = 'file';
+        filePicker.setAttribute('accept', 'image/*');
+        filePicker.style.display = 'none';
+        filePicker.addEventListener('change', (event) => {
+            // Handle the file here
+            console.log(event);
```
dkrasner, line 120: might want to remove this console.log but not important

dkrasner:
@darth-cheney this is nice. I loaded a bunch of .jpg's - all good (although our halo is still funky, but maybe you solved that in the #116 PR)

i was not able to load svg's. Maybe it had to do with the svg I have but I tried a couple and kept getting this error: 

image

I can't upload one of the svg's in question (not supported git format) but I will post it in slack

darth-cheney:
@dkrasner good catch. I just pushed a fix that should deal with the SVG stuff. I also added a new example SVG to the images dir.

PR #121, opened 2021-02-24
String Concatenation

darth-cheney:
## What ##
This PR (finally!) implements a first pass at string concatenation.
  
## Implementation ##
You can concatenate any two Expressions using the `&` symbol. The expressions on either side of the symbol **will be coerced into strings** and concatenated into a resulting string. This has the advantage of easily turning numerical values into strings that can be displayed.
  
Because any variable is also a Factor, and any Factor is an Expression, you can concatenate variables as well. These will also be coerced into strings first.
  
## Examples ##
```simpletalk
on click
    answer "Hello, " & "world!"
end click
```

```simpletalk
on click
    put "world!" into mySuffix
    answer "Hello, " & mySuffix
end click
```

```simpletalk
on click
    answer "You clicked " & 2.4 & " times on average"
end click
```

```simpletalk
on click
    put 4 into numCycles
    answer "You clicked & (numCycles * 10) & " times"
end click
```
  
## Tests ##
I have added [grammar tests](https://github.com/UnitedLexCorp/SimpleTalk/blob/eric-string-concatenation/js/ohm/tests/test-string-grammar.js) and [preload/integration tests](https://github.com/UnitedLexCorp/SimpleTalk/blob/eric-string-concatenation/js/objects/tests/test-string-concatenation-with-preload.js).

### DIFF
--- a/js/objects/tests/test-string-concatenation-with-preload.js
+++ b/js/objects/tests/test-string-concatenation-with-preload.js
@@ -1,0 +1,64 @@
+import chai from 'chai';
+import ohm from 'ohm-js';
+const assert = chai.assert;
+const expect = chai.expect;
+
+import interpreterSemantics from '../../ohm/interpreter-semantics.js';
+
+let testLanguageGrammar = ohm.grammar(window.grammar);
+
+describe("String Concatenation Interpreter Tests", () => {
+    let semantics;
+    before(() => {
+        semantics = testLanguageGrammar.createSemantics();
+        semantics.addOperation(
+            'interpret',
+            interpreterSemantics(null, window.System)
+        );
+    });
+
+    it("Can concatenate two string literals", () => {
+        let input = `"Hello, " & "world!"`;
+        let expected = "Hello, world!";
+        let matchObject = testLanguageGrammar.match(input, "Expression");
+        assert.isTrue(matchObject.succeeded());
+        let result = semantics(matchObject).interpret();
+        assert.equal(expected, result);
+    });
+
+    it("Can concatenate an integer literal and a string literal", () => {
+        let input = `-2 & " times!"`;
+        let expected = "-2 times!";
+        let matchObject = testLanguageGrammar.match(input, "Expression");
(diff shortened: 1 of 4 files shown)


dkrasner:
I can't get a basic example to work, no idea why (I opened a new tab, head is at  72b2487 which is your last commit here). Here is one that fails
```
on click
   answer "ok " & "or not ok?"
end click
```

image


I can see the tests passing, am I missing something (I did do a basic `answer "ok"` just to make sure the script works and it's not failing due to some hidden chars or whatnot)

js/objects/tests/test-string-concatenation-with-preload.js
```diff
@@ -0,0 +1,64 @@
+import chai from 'chai';
+import ohm from 'ohm-js';
+const assert = chai.assert;
+const expect = chai.expect;
+
+import interpreterSemantics from '../../ohm/interpreter-semantics.js';
+
+let testLanguageGrammar = ohm.grammar(window.grammar);
+
+describe("String Concatenation Interpreter Tests", () => {
```
dkrasner, line 10: none of these test and entire handler like 
```
on click
   answer "ok " & "or not ok?"
end click
```
which in theory should be fine since they are unit tests ... but could explain the fact that i am seeing the tests pass but the above handler fail

darth-cheney:
Aha, this is because the `answer` rule in the grammar only accepts a string literal or variable name. We need it to accept whole expressions.

dkrasner:
ok I was just confused b/c your examples int he PR do use the answer command. Did this work in some other way, or just those examples were never run?

darth-cheney:
No I was mucking with it here and never even committed the change.

dkrasner:
cool so playing around it works

just note the example above:

```
on click
    put 4 into numCycles
    answer "You clicked & (numCycles * 10) & " times"
end click
```

is missing a `"` after "clicked" which through me off for a few mins

dkrasner:
good to merge as far as i am concerned

PR #124, opened 2021-02-25
Reading Property Values in Scripts

darth-cheney:
## What ##
This PR implements a first pass at allowing users to read and use Part Property values in scripts. This ability did not exist until now -- we could set property values, but not "read" them.
  
## Implementation ##
The grammar change is simple enough. The pattern is, generally, `the "<propName>" of <ObjectSpecifier>`. There is an alternate form where you can drop the `of <ObjectSpecifier>` portion and the Part will implicitly be the one invoking the script handler.
 
Note also that this new rule `PropertyValue`, is also a `Factor`, meaning it can be used in Expressions or any other rule that can accept expressions.
 
## Examples ##
```simpletalk
on click
    -- Answer the "name" property value of the button
    answer the "name"
    -- Also answer the "name" property value of the button,
    -- but using a specifier
    answer the "name" of this button
    -- And again, but even more roundabout
    answer the "name" of button 2 of third card of current stack
    -- Add to a property and set it again all in one line
    -- (will work after rebase)
    set the "top" of this button to (the "top" of this button + 5)
end click
```
  
## Tests ##
There are grammar and integration tests

### DIFF
--- a/js/ohm/simpletalk.ohm
+++ b/js/ohm/simpletalk.ohm
@@ -66,6 +66,7 @@
         Factor (a factor)
                = "(" Expression ")" --parenFactor
                | "not" Expression --notFactor
+               | PropertyValue
                | anyLiteral
                | variableName
 
@@ -164,6 +165,12 @@
         CommandArgumentList
                 = NonemptyListOf<Expression, ",">
 
+
+        // Reading property values of parts is the same
+        // as asking for such value on an ObjectSpecifier
+        PropertyValue
+                = "the" propertyName "of" ObjectSpecifier --withSpecifier
+                | "the" propertyName --withoutSpecifier
 
         // A "terminal" specifier is one that cannot have
         // "of" attached to the end of it because it's the
@@ -284,6 +291,11 @@
 	// closing quotes
 	stringLiteral
 		= quoteMark (~lineTerminator ~quoteMark any)* quoteMark
+
+        // For now, a property name is wrapped in quotes, ie it
+        // is the same as a stringLiteral
+        propertyName
+                = stringLiteral
 
 	quoteMark = "\""
 
--- a/js/ohm/simpletalk.ohm
+++ b/js/ohm/simpletalk.ohm
@@ -211,11 +211,11 @@
 		| "go to" systemObject objectId --goToByReference
 		| "answer" Expression --answer
 		| "delete" "this"? systemObject objectId? --deleteModel
-    | "set" Expression "to" Expression InClause? --setProperty
+                | "set" Expression "to" Expression InClause? --setProperty
 		| "add" systemObject stringLiteral? "to"? ObjectSpecifier --addModel
 		| "put" Expression "into" "global"? variableName --putVariable
 		| "ask" anyLiteral --ask
-    | "tell" ObjectSpecifier "to" Command --tellCommand
+                | "tell" ObjectSpecifier "to" Command --tellCommand
 		| messageName CommandArgumentList? --arbitraryCommand
 
 	CommentLines
--- a/js/ohm/simpletalk.ohm
+++ b/js/ohm/simpletalk.ohm
@@ -211,7 +211,7 @@
 		| "go to" systemObject objectId --goToByReference
 		| "answer" Expression --answer
 		| "delete" "this"? systemObject objectId? --deleteModel
-                | "set" Expression "to" Expression InClause? --setProperty
+                | "set" propertyName "to" Expression InClause? --setProperty
 		| "add" systemObject stringLiteral? "to"? ObjectSpecifier --addModel
 		| "put" Expression "into" "global"? variableName --putVariable
 		| "ask" anyLiteral --ask
--- a/js/objects/tests/test-property-value-with-preload.js
+++ b/js/objects/tests/test-property-value-with-preload.js
@@ -1,0 +1,122 @@
+/**
+ * PropertyValue semantic rule tests
+ */
+import chai from 'chai';
+const assert = chai.assert;
+const expect = chai.expect;
+
+let currentCardModel;
+let buttonModel;
+
+function compileButtonScript(aScript){
+    let msg = {
+        type: 'compile',
+        codeString: aScript,
+        targetId: buttonModel.id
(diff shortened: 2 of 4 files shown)


ApproximateIdentity:
I've been trying to test this, but am running into small issues. For example this click handler always results in an error:

```
on click
	-- comment
	answer "hello"
end click
```

```
Uncaught missingSemanticAction: Missing semantic action for 'comment' in operation 'interpret'
Action stack (most recent call last):
  interpret > StatementLine
  interpret > comment
```

I went back in the git history and checked it on older versions, but keep getting the same thing. Were comments implemented? Am I messing up something basic here?

darth-cheney:
@ApproximateIdentity Yeah sorry about that. This has been fixed on my serialization PR branch but it hasn't been merged in yet. I keep assuming it's there.

dkrasner:
This looks good. 

One thing is that we have a discrepancy about setting and getting properties in object. [We set](https://github.com/UnitedLexCorp/SimpleTalk/blob/9e5af83e50bbb5058ed6f1c5bcb2b964313beb40/js/ohm/simpletalk.ohm#L215) with an "inClause" i.e. "in [obj specifer]" but we get with an "of [obj specifier]" 

As a matter of fact i don't think the example above can work @darth-cheney since it's grammatically incorrect:
```
set the "top" of this button to (the "top" of this button + 5)
```

If it works that's weird... 

Either way we should commit to one option here. Easier is to use "in" since that's what we have no; maybe more natural is to use "of"

```
set "background-color" to "red" of this button
```

unclear...

darth-cheney:
@dkrasner I have rebased on master and also pushed a commit that updates the `set` command grammar rule to take any Expression as the value to set. We definitely want to be able to do that. Now it should be grammatically correct.

dkrasner:
@darth-cheney you want the name to also be a variable, i have something similar in the [prop changed branchd](https://github.com/UnitedLexCorp/SimpleTalk/tree/daniel-command-propChanged)

but i don't see how this addresses the "in" vs "of" discrepancy above

darth-cheney:
> but i don't see how this addresses the "in" vs "of" discrepancy above
  
I don't necessarily see this as a discrepancy, though we can talk about what would be better. Perhaps for accessing we can use "the <propName> from <objectSpecificer>" or something like that. Either way I think we will eventually obviate this problem by having "put into" be for more than variable assignment and also handle prop setting. That's for down the road however.

dkrasner:
i don't think we should have different ways to specify object properties, in object, of object, from object. Just stick to one

regardless

```
on click
   set "background-color" to "black" of first button in current card
end click
```
does not work, as expected. 

And neither does the script example above, but I can't tell why since the grammar error somehow doesn't invoke the error handler? did something break there


at least we should fix the example in the PR to work in the PR

ApproximateIdentity:
Yeah I have the same issue as Daniel. I.e. if I save the following in the script:

```
on click
    answer the "name"
    answer the "name" of this button
    answer the "name" of button 2 of third card of current stack
    set the "top" of this button to (the "top" of this button + 5)
end click
```

then I get the following error back:

```
on click
    answer the "name"
    answer the "name" of this button
    answer the "name" of button 2 of third card of current stack
    set the "top" of this button to (the "top" of this button + 5) --<<<[Expected:"end"; ruleName: "undefined"]
end click
```

If I keep the comments in (as in your example), I get the following in the console:

```
Uncaught TypeError: match is null
    handleGrammarMatchError errorHandler.js:33
    handle errorHandler.js:13
    receiveError Part.js:337
    receiveMessage Part.js:330
    sendMessage System.js:266
    sendMessage Part.js:316
    compile System.js:328
    receiveMessage System.js:277
    sendMessage System.js:266
    sendMessage Part.js:316
    scriptChanged PartView.js:159
    primHandlePropChange PartView.js:151
    receiveMessage PartView.js:134
    sendMessage System.js:266
    sendMessage Part.js:316
    propertyChanged Part.js:496
    propertyChanged Part.js:495
    setValue PartProperties.js:42
    setPropertyNamed PartProperties.js:208
    click System.js:1263
    receiveCmd Part.js:356
    receiveMessage Part.js:324
    sendMessage System.js:266
    sendMessage Part.js:316
    onClick ButtonView.js:109
```

I agree with Daniel that the example should be working before the merge.

darth-cheney:
@ApproximateIdentity and @dkrasner --
  
First, I've rebased and pushed some slight changes that should make what I describe below work.
  
Second, you are correct -- the final example was not grammatically correct. This is because I was saying `set the "top"` instead of `set "top"`. We have used "the" to help distinguish between property _values_ (rather than their names). At the moment, we cannot query the `PropertyValue` rule (`the "something" of`) to get its property name only. The `interpret()` semantics is to look it up and return the current value. That's it.
  
The correct command to use per the example is one of these two, which are equivalent:
```simpletalk
set "top" to (the "top" of this button + 5)
set "top" to (the "top" of this button + 5) in this button
```

ApproximateIdentity:
This seems to be working for me now.

js/ohm/simpletalk.ohm
```diff
@@ -211,7 +211,7 @@
 		| "go to" systemObject objectId --goToByReference
 		| "answer" Expression --answer
 		| "delete" "this"? systemObject objectId? --deleteModel
-                | "set" Expression "to" Expression InClause? --setProperty
+                | "set" propertyName "to" Expression InClause? --setProperty
```
dkrasner, line 214: won't this prevent from doing things like 

```
put "-color" into var
set "text"&var to "blue"
```

I think we want to be able to have this sort of dynamics

darth-cheney: You are correct -- it prevents that. I was having some arcane issue dealing with expressions in that position though, and I figured we could just come back to it when we need it. I can open an issue also

dkrasner:
@darth-cheney merge this in?

dkrasner:
@darth-cheney i deleted this branch. I think we should delete branches that are merged

sometimes i go through and delete yours but i am not consistent and they accumulate: https://github.com/UnitedLexCorp/SimpleTalk/branches

PR #126, opened 2021-02-26
Eric stepping

dkrasner:
### Main Points ###

Adding a setInterval/stepping property. 

Example:

```
on click
   put 0 int0 global varCounter
   set "stepping" to true
end click

on step
   set "name" to varCounter
   put (varCounter + 1) into global varCounter
end step
```


ApproximateIdentity:
@dkrasner 

I get an error using the bootstrap example. I add a button to card and then try to put the following into the buttons script:

```
on click
   put 0 int0 global varCounter
   set "stepping" to true
end click
```

```
on click
   put 0 int0 global varCounter --<<<[Expected:"end" or a statement list, comments included; ruleName: "StatementList"]
   set "stepping" to true
end click
```

I checked the same thing on all 4 commits in this PR and get the same error for all of them so I guess the example script is buggy.

dkrasner:
yeah @ApproximateIdentity  you have "int0" but should have "into"   "0" -> "O"

PR #127, opened 2021-02-26
command prop changed

dkrasner:
### Main Points ###

After a view receives a `propertyChanged` type message it will [send](https://github.com/UnitedLexCorp/SimpleTalk/compare/daniel-command-propChanged?expand=1#diff-f8a644f39713f7d56b51b944769a8733fceb081884704c2a73b93c232b554560R150) a type `command` `propertyChanged` message to its model/part. This message is set not to delegate, which is a new option for messages handled [here](https://github.com/UnitedLexCorp/SimpleTalk/compare/daniel-command-propChanged?expand=1#diff-efd61be9944474fec7ba0b89d7a931fedae606964b569ff27b2766271c5244b9R387). 

Note: the `propertyChanged` command is sent **regardless** of whether the view subscribes to the property being changed. This is necessary because a given property might (dynamically) alter another property which the view subscribes to. Our StyleProps are a good example. This allows a separation + translation layer between SimpleTalk and whatever is necessary to property render the thing. For example, "text-color" is a SimpleTalk property but it's corresponding CSS is "color" which is set by the `cssStyler` in the `cssStyle` property. The latter is what a view actually subscribes to. 

Anyhow the changes here are all pretty simple and I've added tests. 

++ 
Fun example:

image

script for button "Controller"
```
on propertyChanged name, value
   if name == "background-color" then setTarget name, value
   if name == "left" then setTarget name, value
   if name == "text-color" then loadImage
end propertyChanged

on setTarget name, value
   set name to value in button "Target"
end setTarget 

on loadImage
   tell first image of current card to loadImageFrom "https://source.unsplash.com/random"
end loadImage
```

### DIFF
--- a/js/objects/parts/Part.js
+++ b/js/objects/parts/Part.js
@@ -376,14 +376,17 @@ class Part {
             var result = boundHandler(aMessage.senders, ...aMessage.args);
             window.System.executionStack.pop();
             return result;
-            
         }
 
         // Otherwise, we have no handler for
-        // it, so we delegate along the
+        // it. Unless the message indicates shouldNotDelegate
+        // we delegate along the
         // message delegation chain. It is up
         // to Parts to properly implement delegation
         // for themselves!
+        if(aMessage.shouldNotDelegate){
+            return aMessage;
+        }
         return this.delegateMessage(aMessage);
     }
 
--- a/js/objects/properties/PartProperties.js
+++ b/js/objects/properties/PartProperties.js
@@ -136,6 +136,12 @@ class StyleProperty extends BasicProperty {
 
             // set my value as well
             this._value = val;
+            if(notify){
+                owner.propertyChanged(
+                    this.name,
+                    val
+                );
+            }
         }
     }
 };
--- a/js/objects/tests/test-view-prop-handling-preload.js
+++ b/js/objects/tests/test-view-prop-handling-preload.js
@@ -70,6 +70,65 @@ describe('Test use of #onPropChange registration', () => {
         );
     });
 });
+describe('Test #onPropChange command delegation', () => {
+    var buttonModel;
+    let buttonView;
+    var buttonPropertyName;
+    var buttonPropertyValue;
+    var ownerPropertyName;
+    var ownerPropertyValue;
+    before('Can create View/Model pair', () => {
+        let stackEl = document.querySelector('.current-stack');
+        let cardEl = stackEl.querySelector('st-card');
+        let msg = {
+            type: "command",
+            commandName: "newModel",
+            args: ["button", cardEl.model.id]
+        };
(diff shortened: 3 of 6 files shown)


ApproximateIdentity:
I went through this and all the interaction works as expected. I.e. I went through your example and it all is working as you describe. I didn't jump into the code that much. The change seem relatively small so obviously the main changes come from the overall interaction with the system which I'm not totally following. But ignoring my ignorance of the whole setup, everything else is clean. I guess @darth-cheney can speak more to that.

One thing I noticed though @dkrasner is that there is one test that is failing:

```
$ npm run test

[ ... cut out most of them ... ]

  newModel tests                                                                                                                                                                                                   
    ✓ System has loaded                                                                                                                                                                                            
    ✓ The current card does not have any button models or views                                                                                                                                                    
    ✓ Can send newModel message without error (add button to current card)                                                                                                                                         
    ✓ Can send newModel message without error (add image to current card)                                                                                                                                          
    - Can send newModel message without error (add default image)                                                                                                                                                  
    ✓ Can send newModel message without error (add image to current card without id)                                                                                                                               
    ✓ Can send newModel message without error (add image to current card, in context)                                                                                                                              
    ✓ Current card view has a child button view                                                                                                                                                                    
    ✓ Can send newModel message in a context (without ownerId) without error (add button to current card)                                                                                                          
    ✓ Current card view has a second child button view                                                                                                                                                             
    1) There should be a serialization for the new button                                                                                                                                                          
    ✓ Can send newModel message in a context (without targetId, current)                                                                                                                                           
    ✓ Can send newModel message in a context (without targetId, this)                                                                                                                                              
    ✓ Can send newModel message in a context (without targetId, targetType nor context)                                                                                                                            
                                                    
  copyModel tests                                   
    ✓ Can send copyModel message    
                                                    
  deleteModel tests                                 
    ✓ Can send deleteModel message without error (delete the button by id)
    ✓ Can send deleteModel message without error (delete the button without id, self referentially)
    ✓ Current card should delete button subpart by Id                     
    ✓ Current CardView should not have any direct ButtonView children after it is deleted
    ✓ Current card should delete button subpart wihout Id, self-referentially
                                                                                                         
                                                                                                         
  18 passing (125ms)                                
  1 pending                                         
  1 failing                                                                                              
                                                                                                         
  1) newModel tests                                                                                      
       There should be a serialization for the new button:
     SyntaxError: Unexpected token u in JSON at position 0                          
      at JSON.parse (<anonymous>)                                                                        
      at Context.<anonymous> (js/objects/tests/test-system-model-add-remove-preload.js:132:34)
      at processImmediate (node:internal/timers:463:21)
```

dkrasner:
@ApproximateIdentity yeah Eric knows about this, coming from the serialization pr

PR #128, opened 2021-03-01
Serialization IDs

darth-cheney:
## What ##
This PR ensures that IDs are preserved and correctly loaded again across the serialization / deserialization lifecycle.
  
## Implementation ##
The previous issue was that some Parts, like the Stack, will create an initial Card as a subpart of themselves when constructed. This was throwing off our `idMaker` module, which for the moment is a dumb incrementer. 
  
The strategy employed here does the following:
1. Ensures that each Part knows how to deserialize itself, via a class method `fromSerialized`. This method should return a Part instance. Note that WorldStack and Window have special behavior for this method. Otherwise all other parts use the Part.js base implementation
2. Resets the `idMaker`'s count to be the maximum ID value once all Parts have been deserialized. This ensures that the "next" created Part by the user will have the correct next ID, and that there are not clashes.
  
Additionally, I have adjusted some methods here and there to make them easier to test.
  
## Tests ##
There is now a [preload test file for serialization/deseralization](https://github.com/UnitedLexCorp/SimpleTalk/blob/eric-serialization-ids/js/objects/tests/test-serialization-with-preload.js). We test both the JSON parts and the saving of the HTML document. They are currently passing fine.

### DIFF
--- a/js/objects/System.js
+++ b/js/objects/System.js
@@ -35,6 +35,8 @@ import ohm from 'ohm-js';
 import interpreterSemantics from '../ohm/interpreter-semantics.js';
 import {ExecutionStack, ActivationContext} from './ExecutionStack.js';
 
+import idMaker from './utils/idMaker.js';
+
 const video = document.createElement('video');
 const canvas = document.createElement('canvas');
 var handDetectionModel = null;
@@ -739,7 +741,7 @@ const System = {
         if(!serializationEl){
             throw new Error(`No serialization found for this page`);
         }
-        let deserializedInfo = JSON.parse(serializationEl.innerText);
+        let deserializedInfo = JSON.parse(serializationEl.textContent);
 
         // Start from the WorldStack and recursively
         // create new Parts/Views from the deserialized
@@ -783,6 +785,19 @@ const System = {
                 );
             }
         });
+
+        // Finally, we reset the idMaker to start its
+        // count at the highest current id
+        let numericIds = Object.keys(System.partsById).filter(id => {
+            return id !== 'world';
+        }).map(id => {
+            return parseInt(id);
+        });
+        let currentId = Math.max(...numericIds);
+        idMaker.count = currentId;
+
+        // And serialize
+        this.serialize();
     },
 
     serializePart: function(aPart, aDict){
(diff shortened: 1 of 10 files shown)


ApproximateIdentity:
@darth-cheney I can't load the bootstrap.html example page on this branch. I get the following error in the console (which I haven't been able to understand):

```
Uncaught ReferenceError: i is not defined
    js idMaker.js:1
    Webpack 9
        __webpack_require__
        js
        __webpack_require__
        js
        __webpack_require__
        js
        __webpack_require__
        <anonymous>
        <anonymous>
```

Does the bootstrap page load for you?

darth-cheney:
@ApproximateIdentity Whoops. Somehow a small typo got into that commit at the end. I've pushed a fix and it should be working now

dkrasner:
@darth-cheney this all looks good to me. 

A minor point. It seems like there is no need for [subpart ids](https://github.com/UnitedLexCorp/SimpleTalk/blob/1a22d3f53cdc536070b2b48f17a1b4647378022c/js/objects/parts/Part.js#L516) in the serialization. They don't provide any new information. Should we remove these?

darth-cheney:
@dkrasner The subpart ids are used [here](https://github.com/UnitedLexCorp/SimpleTalk/blob/1a22d3f53cdc536070b2b48f17a1b4647378022c/js/objects/System.js#L847) during deserialization. They are also important to keep because they specify the current order in the owner.

ApproximateIdentity:
@darth-cheney There seems to be some sort of bug with either the serialization or the saving. If you open the bootstrap.html page, then save it, then move that the bootstrap.html, then open it again and save it as (say bootstrap2.html), then you don't get back the original file:

```
$ diff bootstreap.html bootstrap.html 
509c509
< <a style="display: none;"></a><a style="display: none;"></a></body></html>
\ No newline at end of file
---
> <a style="display: none;"></a></body></html>
\ No newline at end of file
```

Basically it looks like `<a style="display: none;"></a>` gets appended even if it's already there.

On a secondary note, the checked in bootstrap.html file has `<!DOCTYPE html><html xmlns="http://www.w3.org/1999/xhtml"><head>` at the top, while the saved version has

```
<!DOCTYPE html>
<html xmlns="http://www.w3.org/1999/xhtml"><head>
```

I think we should fix this by opening the current version, saving it, and then checking in that. This way opening the checked in file and saving it will produce the same file (after fixing the issue with `<a style="display: none;"></a>` that I talked about ealier in this comment of course).

darth-cheney:
@ApproximateIdentity Good catch on the anchor tag thing. I've just pushed a fix for in the most recent commit.
  
As for the whitespace formatting, I'm not sure what to do. I am using the [built-in XMLSerializer](https://developer.mozilla.org/en-US/docs/Web/API/XMLSerializer) for the DOM. As far as I can tell there isn't any API for controlling how whitespace is handled.

ApproximateIdentity:
@darth-cheney Pull down my last commit and test the following:

1. Open the bootstrap example.
2. Save the html.
3. Copy the saved html to the bootstrap.html file (i.e. overwrite the current file)

Do you get the same thing back? I.e. is this process a noop? This is what I meant before. For me this is now a noop.

darth-cheney:
@ApproximateIdentity Now trying saving the bootstrap using Chrome. On my end it does the whitespace differently (the previous way)

ApproximateIdentity:
Out of curiosity what happens if you use firefox? Is this mainly a firefox/chrome issue or is there more variation than that? E.g. maybe mac/linux differ as well. (I sort of assume Windows would differ.)

darth-cheney:
When I use Firefox I don't have a diff between the original and the saved version (after your commit of course). So it seems like the bootstrap you pushed is the "Firefox format" and the original one I had was the "Chrome format"

ApproximateIdentity:
Okay cool good to know. I don't know if it makes sense to mess with this too much more at this level. Maybe better would be to use a smarter diff algorithm. I.e. basically something that would check if the actual html structures are the same. One possibility would be to only look at the tree structure by ignoring all formatting between elements. I don't mean to remove the files, but to have a binary `smartdiff` that would basically be the same as `diff` except that it would recognize if two html files are the same even if they have slightly different formatting.

Anyway this isn't all that important just something to think about. My personal opinion is that this should be merged.

Issue #130, opened 2021-03-04
Toolbox and visibility


dkrasner:
###Main issue###

Toolbox is only visible on the first card of the stack. This is simply because it is the 2nd child of the current stack, after the first card, and the DOM puts it to the forefront. However if we move to the next card, where Toolbox is now a younger sibling, it is covered over the it. 

I see two obvious options:
* set the z-index of windows with name "Toolbox" to something high
* make sure that cards are added before the Toolbox, i.e. that it is the last child of Stack

PR #131, opened 2021-03-05
Audio

dkrasner:
### Main Points ###

This PR adds an Audio part and view. 

These as you can see are pretty minimal. The current AudioView does not display itself (display: none), although there is nothing stopping us from giving it a visual presence in the future. The Audio part takes a few basic private commands: `loadFromSource`, `play`, `pause` and `stop` (which resets play to the beginning). 

The `play` command which sets the property "play" to `true`, first [checks the ready state](https://github.com/UnitedLexCorp/SimpleTalk/compare/daniel-audio?expand=1#diff-6303ddf2c9a19e094464b179ca1510f080fd538ee9ea41fec0de5e8063e72323R74) of the audio element and only plays if it is ready. Otherwise it sends alert. In the future this should be some global message, maybe an error message, like "ResourceNotReady."

#### Example ####
```
add audio to current card
tell first audio of current card to set "name" to "theDuke"
tell audio "theDuke" of current card to loadAudioFromSource "https://ia801506.us.archive.org/31/items/lp_harlem-new-world-acoming-the-golden-broo_duke-ellington-cincinnati-symphony-orchest/disc1/01.01.%20New%20World%20A%27Coming.mp3"

tell audio "theDuke" of current card to play
tell audio "theDuke" of current card to pause
tell audio "theDuke" of current card to stop
```

(these private commands at the moment are simply conveniences, since they simply change underlying basic props, i.e. you can do `...set "play" to false` to pause the thing. But i think it's nice to have them, and they are more intuitive than the media html element API which, for example, doesn't have a `.stop()` asking you to use the `.load()` to achieve the samething)

Note: I [set the z-index to 100]( https://github.com/UnitedLexCorp/SimpleTalk/compare/daniel-audio?expand=1#diff-c753a1396e7b4b55330c513675b157cc4aed6cb4508b6f00831bf44d20183c4dR113) for toolbox due to this issue #130 which I will work on separately.


ApproximateIdentity:
Each time I press stop, a 1.25MB fragment of the original song seems to be downloaded. I can see it in the network tab at each click and I checked my actual system resources to double-check that yes it is definitely redownloading each time. I don't have the cache disabled so that's not it. Is this happening for the rest of you?

ApproximateIdentity:
I like how you can get multiple streams going simultaneously. So sounds can play over each other out of the box allowing e.g. fading between songs and notifications playing while other sounds are playing.

darth-cheney (approved):
This is cool!
  
Just a couple of things:
We should add the `audio` part to certain tests, like [here](https://github.com/UnitedLexCorp/SimpleTalk/blob/daniel-audio/js/ohm/tests/test-object-specifier-grammar.js#L23), to make sure it's covered.
  
Another thing to consider is this -- do we want to have Parts that don't have corresponding views? Something worth discussing I think.
  
The final note here is only slightly related: we might want to have more of the functionality of the WebAudio API built into SimpleTalk, maybe even at the System level. Hypercard had ways of composing notes and playing a few stock synth instruments in addition to actual sound clips. Of course, something like that is totally not a priority at the moment (but makes for compelling demos)

dkrasner:
> I like how you can get multiple streams going simultaneously. So sounds can play over each other out of the box allowing e.g. fading between songs and notifications playing while other sounds are playing.

yeah didn't think of that, but clearly should be an easy thing to do if things are setup right

dkrasner:
> We should add the `audio` part to certain tests, like [here](https://github.com/UnitedLexCorp/SimpleTalk/blob/daniel-audio/js/ohm/tests/test-object-specifier-grammar.js#L23), to make sure it's covered.
> 
yeah i'll add a few in. Wanted to get your comment first before going test-crazy
> Another thing to consider is this -- do we want to have Parts that don't have corresponding views? Something worth discussing I think.
> 
I think for now keeping our model-view separation is good, and simply having views which don't display (display: none) is a simple way to roll. This way everything we current have in terms of language/interpreter etc function exactly the same way, and it's easy to debug (b/c we can always look at the DOM/dev console and see what's what).

The only potentially confusion component is object reference. For us this is not an issue, b/c we can imagine things that are there on the DOM which can't be seen, so doing something like "tell first audio of current card...." makes sense, but to a ST author-user this might not have any tangible reference. But I think that this is something that needs to be worked out in practice with people trying the thing out. 

In addition, having these no-display part views makes it really easy to go in and give them a view. Suppose we decide that an audio should have a view, like a play-icon-button, then this becomes a trivial change.

> The final note here is only slightly related: we might want to have more of the functionality of the WebAudio API built into SimpleTalk, maybe even at the System level. Hypercard had ways of composing notes and playing a few stock synth instruments in addition to actual sound clips. Of course, something like that is totally not a priority at the moment (but makes for compelling demos)

I've been thinking about this, on my "time off"... lots of different option and worth investigating for sure. An audio editor would also be a good thing. One thing would be amazing is both "sheet music (programmable nomenclature) -> audio" && "audio -> sheet music." But the latter translation is inherently non-deterministic (depends on the key and time signature you choose), so I am not sure what the current/modern standard is for such translation. Anyhow i'll look into this after `field` is where we want it. More of a priority.

dkrasner:
this could be a decent place to start: https://tonejs.github.io/

It's a library recommended here https://developer.mozilla.org/en-US/docs/Web/API/Web_Audio_API/Advanced_techniques

the API looks pretty easy and we could immediately bootstrap a few examples, wrapping their API in ST commands/prop changes...

dkrasner:
> Each time I press stop, a 1.25MB fragment of the original song seems to be downloaded. I can see it in the network tab at each click and I checked my actual system resources to double-check that yes it is definitely redownloading each time. I don't have the cache disabled so that's not it. Is this happening for the rest of you?
> 
> edit: I changed this to play song locally (i.e. threw a random mp3 in `SimpleTalk/music/song.mp3` and then loaded `http://localhost:8000/music/song.mp3`) and that is still redownloading each time I press stop, but this time it is _cached_. So maybe the reason the other one is not cached is because their server sends headers not to cache or something. Just some speculation. But either way I guess we shouldn't be dependent on that

my guess is that we need more control of playing, pausing and stopping. Current `stop` calls `audio.load()` which is the only way I saw to actually reset play to the beginning. But in the future we'll probably move to a richer web audio interface (https://developer.mozilla.org/en-US/docs/Web/API/Web_Audio_API)

ApproximateIdentity:
> my guess is that we need more control of playing, pausing and stopping. Current `stop` calls `audio.load()` which is the only way I saw to actually reset play to the beginning. But in the future we'll probably move to a richer web audio interface (https://developer.mozilla.org/en-US/docs/Web/API/Web_Audio_API)

Yeah we can just open an issue and look at it down the road.

dkrasner:
@ApproximateIdentity done #137 

going to merge this in

PR #133, opened 2021-03-10
Layouts v2

### LINKED ISSUE #115: "Current-ness" as Property
## What ##
As it stands now, the "current-ness" of a Stack (in the World) or a Card (in a Stack) is determined solely by adding a class to the corresponding view's element. This is also how we navigate through Cards and Stacks.
  
We need to change this -- the current-ness should be a property on either the Card or the Stack in question, and the views/System should react to that property changing.
  
## Options ##
There are two ways to go, and it's unclear which is better.
  
In the first method, a Card changes its "current" boolean property to true. It would be a dynamic prop that then goes into its parent Stack and sets all other Cards "current" to false.
  
In the second method, the Stack itself keeps track of which of its Card subparts is the current one, as some property on itself. This is that changes and views react to that change.

darth-cheney:
## What ##
This PR re-implements layouts and layout related properties across relevant parts. 
  
Additionally, it implements some fixes for tangential issues that I discovered in the course of creating the examples stack.  
  
## Live Examples ##
Before diving into a description of the changes, it's important to know that I have created a live examples snapshot that allows you to play with and manipulate many of the layout related properties. It is located [here](https://github.com/UnitedLexCorp/SimpleTalk/blob/eric-layouts-new/js/objects/examples/layout-examples.html).
  
Please comment on this PR if you have any questions or notice anything you think is not working correctly in these examples. All of them were created entirely using SimpleTalk scripting and Parts -- there was never anything hard coded.
  
### Style Props Adjustment ###
Before adding layout specific styling properties, I needed to ensure that certain kinds of objects always had basic sets of styling properties that use the `styler` pattern. To achieve this, I created a new [utility file](https://github.com/UnitedLexCorp/SimpleTalk/blob/eric-layouts-new/js/objects/utils/styleProperties.js) with some simple methods that Parts can now import and will  call in their constructors.
  
Note also that many of the actual layout related properties are not technically style properties -- they are either basic ones or dynamic ones. This is because we use CSS classes and the core stylesheet to more easily work with layouts, which can be complicated, and that requires having views react to individual property changes. These change handlers have been added to the base PartView.js and only affect concerned Parts.
  
### About The Layout Structure ###
In general, there are only two types of layouts, which are set using the `layout` part property:
1. `strict` -- Subparts are treated as if they should be absolutely positioned, and their `top` and `left` part properties will indicate offsets from the parent Part's top left corner;
2. `list` -- Subparts will be algorithmically aligned and distributed in the parent part according to additional related part properties. Under the hood, this is a simple wrapper for flexbox.
  
#### Kinds of `list` layouts ####
The part property `list-direction` specifies one of two possible list types, `row` (left to right) or `column` (top to bottom). If the `layout` property of the part is set to `strict`, the `list-direction` property, along with other list layout related properties, will have no effect.
  
There are other list related properties that will have additional effects and I will direct you to the example stack to see them in action. But in short, these include:
* `list-alignment` -- How subparts should be aligned along the dominant axis
* `list-distribution` -- How subparts should be distributed throughout the breadth of the dominant axis in the parent Part
* `list-wrapping` -- How to handle cases where the parent Part overflows when in list mode

### Who can use `layout`? ###
For now, there are only two kinds of parts that can have the `layout` and `list-` related properties: `Card` and  a new part we call `Area`.  This part is simply a grouping of other subparts that can appear in a card. It is the primary way -- aside from setting layout properties on a Card itself -- to create complex, dynamic layouts within a Card. Most of the examples in the layout example stack make use of Areas.
  
### Resizing ###
An important set of properties of the subparts of list-style layouts exists which tell us how each subpart should try to fill space inside of its container. They are `horizontal-resizing` and `vertical-resizing`. They take the following possible values:
* `rigid` - Adhere to the `width` (for horizontal) or `height` (for vertical) amount
* `space-fill` - Attempt to fill as much space as possible along the given direction. In the case of list layouts, this is akin to flex, so multiple subparts with `space-fill` will each divide the amount of free space between each other.
* `shrink-wrap` - Only take up as much space along the given direction as is required to fix the subpart's own subpart contents.
  
For the moment it is *unclear* what the effect of `space-fill` should be when used on subparts that are children of a `strict` layout part. This is because coordinates are relative. We could say it means the width matches the parent, but if the subpart then has positive `top` or `left` values, it will overflow the parent. For the moment I have simply said this configuration has no effect in `strict` layouts. We might want to add  a `match-parent` option that will take its place in these situations.
  
### `if-then-else` blocks ###
In short: these were not working properly, and our tests were missing them due to an arcane grammatical situation. I have fixed these and it now works, and they are used plenty in the layout examples stack
  
### Current-ness of Stacks and Cards ###
A long-standing issue with deserialization and determining how and when stacks or cards should be considered "current" finally came to a head (see Issue #115), necessitating that I finally refactor that whole situation. Instead of purely being determined by a CSS class on the respective views, "current-ness" is now a dynamic part property on Cards and Stacks, and the views respond to changes accordingly.
  
The key thing to know about the implementation is that by setting a Card or Stack's `current` property to true, it will look up its siblings and set their (private) current property to false first. This triggers the appropriate changes in the views.
  
These changed required updating how Stack/Card navigation works (Part-side now, as opposed to View-side previously) and updating the tests accordingly. They are all now passing and I think we will be able to close #115 unless you all notice other bugs.
  
### Font Stuff ###
I have added a new default font called Crimson Pro, which I think looks nice for typography and display. We don't have to keep it, but I liked it for the example stack.
  
## Closes ##
#115


ApproximateIdentity:
I just started playing with the main layouts example and found something weird. If I click and drag the buttons in the box below "Basic Examples", they get duplicated a little up and to the left for each button. It seems they only get duplicated once and the duplicate buttons seem to work just as their counterparts. Is this intentional?

darth-cheney:
@ApproximateIdentity I think that has to do with the current "drag and drop"" functionality that has been left in. This is definitely something we need to revisit, as I think we should bypass normal DOM drag and drop actions and completely implement our own system for it. I will open an issue

dkrasner:
@darth-cheney a lot of good changes here, will need more time to go through it

but take a look at my comment in #115 which is not what is implemented:

"
The key thing to know about the implementation is that by setting a Card or Stack's current property to true, it will look up its siblings and set their (private) current property to false first. This triggers the appropriate changes in the views.
"

Although it might be a question of perspective, if we are in a framework where a part is changing properties of itself and all its siblings, it really should be the parent changing the properties of its children/subparts. This is what I would expect in a general parent-children hierarchy. Moreover, the parent will have more information, or higher level view, on what the state of its children is which might influence the next state. 

For example, imagine we want to do a transition from one card to the next, say a fade-out -> in. I would want to do something like `Stack.fadeOutCurrent()` `Stack.fadeInCurrent(card.id)` which should not require any sibling lookups etc

dkrasner:
> ### Who can use `layout`?
> 
> For now, there are only two kinds of parts that can have the `layout` and `list-` related properties: `Card` and a new part we call `Area`. This part is simply a grouping of other subparts that can appear in a card. It is the primary way -- aside from setting layout properties on a Card itself -- to create complex, dynamic layouts within a Card. Most of the examples in the layout example stack make use of Areas.

I prefer `Area` to `Container`  - we can reserve the latter name for something else if ever need one. 

`Window` + `Area` seems like all you ever need 
> ### Resizing
> 
> An important set of properties of the subparts of list-style layouts exists which tell us how each subpart should try to fill space inside of its container. They are `horizontal-resizing` and `vertical-resizing`. They take the following possible values:
> 
>     * `rigid` - Adhere to the `width` (for horizontal) or `height` (for vertical) amount
> 
>     * `space-fill` - Attempt to fill as much space as possible along the given direction. In the case of list layouts, this is akin to flex, so multiple subparts with `space-fill` will each divide the amount of free space between each other.
> 
>     * `shrink-wrap` - Only take up as much space along the given direction as is required to fix the subpart's own subpart contents.
> 
> 
> For the moment it is _unclear_ what the effect of `space-fill` should be when used on subparts that are children of a `strict` layout part. This is because coordinates are relative. We could say it means the width matches the parent, but if the subpart then has positive `top` or `left` values, it will overflow the parent. For the moment I have simply said this configuration has no effect in `strict` layouts. We might want to add a `match-parent` option that will take its place in these situations.
> ##

i think the no-effect is the right way to go. "Rigid" in my mind implies more control from the author-user, setting position, sizing etc

darth-cheney:
@dkrasner The 115 issue seems like a reasonable change. Should we keep the issue open and leave for a separate PR?

dkrasner:
my take is that this is the time to do it, it's pretty fundamental. 

But you've had your head in the sand on this for some days so if you want to hand it off to me I can handle it another PR. If you are happy to do it then lets make the change here.

darth-cheney:
Ok I'll take care of it

dkrasner:
The layout example stack is super fun!

going to start adding issues as I find them:

```
add area to current card
tell second area of current card to set "name" to "new area"
tell area "new area" of current card to add area
```

the last statement there throws an error `Uncaught STVariableReferenceError: Variable area has not been defined`

whereas 

```
tell area "new area" of current card to add area to area "new area" of current card
```
does what you would expect

is that that an interpreter bug or as expected and we need explicit object specification in a `tell` command?

darth-cheney:
> the last statement there throws an error Uncaught STVariableReferenceError: Variable area has not been defined
  
Seems to be a combo grammar/interpreter issue. It's not the full statement that's a problem -- `add area` by itself is not a valid command grammatically, as an ObjectSpeficier is always required at the moment:
```ohm
"add" systemObject stringLiteral? "to"? ObjectSpecifier --addModel
```
I agree though that we should change this and when there is no specifier, the current part in context should implicitly be assumed to be the target.
  
> You see I am selecting the red inner-area's halo, but it's being cropped...
  
We should open an issue because we are going to have to reimplement Halos I think. Right now they are added to the shadow DOM of the targeted views, but that is really screwing up things like cropping (the problem you are seeing) and the earlier issue with transparency. It might be difficult to do another way but we should experiment.

dkrasner (approved):

dkrasner:
@darth-cheney i am good with this. There are certainly issues besides those above, but we can deal with those as time moves fwd. This looks like a bunch of conflicts so if you want to go through them together let me know

PR #136, opened 2021-03-11
Hand recognition grab

ApproximateIdentity:
I think this basically ready to merge. What this allows is simple. If you turn on hand recognition, it creates Lenin's hand. If you then push the hand in, it will "grab" the nearest element in the card and move it around. If you push in again, it will release the element. I will post a video in slack showing the basic flow. I think you guys should basically go through the same flow and verify it works for you guys as well.

The following (at least) needs to be done before the merge:

1. Whatever you guys find that you thinks needs fixing.
2. ~~Rebase to make it works with the recent changes.~~
3. ~~I want to extract all the functionality to a separate utility class.~~

Right now the hand detection is pretty sensitive to lighting. Also it can be hard to see what's happening in the detection for debugging purposes. It might be worth adding a formal video part whose only job is to display the video stream and possibly also the hand detection boxes.

Anyway play around with it. I'm sure there are many bugs that I wont find since I keep doing it in my own way. But I do feel the functionality is there for it to basically be a complete feature (even if the sensitivity of detection jumps around way more than it should).


js/objects/System.js
```diff
@@ -965,8 +974,9 @@ System._commandHandlers['deleteModel'] = function(senders, ...rest){
 };
 //System._commandHandlers['newModel'] = System.newModel;
 System._commandHandlers['newModel'] = function(senders, ...rest){
-    System.newModel(...rest);
+    const model = System.newModel(...rest);
     this.serialize();
+    return model;
```
dkrasner, line 979: `newModel` is a command handler why does it need to return something?

if you want the model you can either query for it or use `System.newModel()` directly

dkrasner:
nice!


maybe we can move everything by the core command handler which starts/stops hand-recognition to a util file location, such as js/object/utils/

and is there a reasonable way to add some tests? For example, placing a couple of buttons and then the hand over them, doing a grab move and making sure the right button is grabbed etc

js/objects/System.js
ApproximateIdentity: I honestly just assumed it was mistake that it wasn't returning something. This is the line calling this function (which I didn't change):

https://github.com/UnitedLexCorp/SimpleTalk/blob/150faac12131f1949d8ebf7b758d8a903d0c6a79/js/objects/System.js#L374

That line seems to want a result and I didn't see any harm doing this. But yeah if you'd rather noy have something returned, I could certainly call `System.newMoedel()` directly instead of sending the message like this.

ApproximateIdentity: I did what you said and changed it to use `System.newModel()` directly in this commit: https://github.com/UnitedLexCorp/SimpleTalk/pull/136/commits/161a8caf7314d259794c9f7afe431a06a7fda9dd

ApproximateIdentity:
> maybe we can move everything by the core command handler which starts/stops hand-recognition to a util file location, such as js/object/utils/

Yup that sounds good. Will do.

> and is there a reasonable way to add some tests? For example, placing a couple of buttons and then the hand over them, doing a grab move and making sure the right button is grabbed etc

Yeah I need to think about that. Certain things can easily be unit tested (like my point to element distance function) and so definitely should do that, but I'd have to think about how best to test the interactions.

dkrasner:
> > 
> 
> Yeah I need to think about that. Certain things can easily be unit tested (like my point to element distance function) and so definitely should do that, but I'd have to think about how best to test the interactions.

don't go too crazy on this. It's not as important as getting the API working and the thing looking/feeling right. But if you see good ways to start testing this then yeah probably a good idea

ApproximateIdentity:
I moved the code to `utils/` and added an empty test file. An issue I have is that `document` doesn't exist in the tests. I looked on the net, but couldn't find a way to mock it so I couldn't get the tests to run. Is there an easy way to do this?

dkrasner:
@ApproximateIdentity that's weird, `document` is definitely there, it's [defined](https://github.com/UnitedLexCorp/SimpleTalk/blob/master/js/objects/tests/preload.js#L14) and we do use it extensively 

are you sure you are running the test with preload?>\

ApproximateIdentity:
Unless you guys disagree, I think this now probably can be merged. I've added basic testing. There should be more tests, but to do it right I should probably clean up the code a little. That's worth doing, but maybe better to avoid that rabbit hole right now. Also I added basic smoothing of the hand itself. It's still limited by the model framerate. There is much to do there, but we might get lucky and get a lot for free by using the big desktop being shipped to me so might be worth trying that first.

Issue #138, opened 2021-03-16
Place current property into Stacks rather than cards


darth-cheney:
Reverse the responsibility and update the navigation methods accordingly. This allows Stack/WorldStack to manage the sibling queries etc.

darth-cheney:
Closed by #140

PR #140, opened 2021-03-16
Issue 138: Currentness

darth-cheney:
## What ##
This PR is a fix for Issue #138, which deals with moving the 'current' property from Cards and Stacks up a level to the owner parts that manage this state, namely WorldStack and Stack.
  
## Implementation ##
The property is still named `current`, but has been moved up a level and now has a different meaning. It refers to the 0-indexed based location of the current item (a card, in the case of Stack or a stack in the case of World) in a filtered list of like-kind items. This was done in anticipation of compatibility with forthcoming `number` properties as well as simplifying the navigation methods.
  
We also had to make minor updates to some System functions and tests, but everything seems to be working.
  
Still each of us should test this live (testing card navigation etc) just to be sure.
  
## Closes ##
#138

### DIFF
--- a/js/objects/System.js
+++ b/js/objects/System.js
@@ -40,7 +40,7 @@ import {ExecutionStack, ActivationContext} from './ExecutionStack.js';
 
 import idMaker from './utils/idMaker.js';
 
-import handInterface from './utils/handInterface.js'
+import handInterface from './utils/handInterface.js';
 
 const System = {
     name: "System",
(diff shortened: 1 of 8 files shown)


dkrasner:
i like it

do we need to update the bootstrap and layout html files?

darth-cheney:
The current files should be fine. They store the IDs of the current card and stack separately at the root of the JSON serialization, and I have updated the reading/writing of that part also

Issue #139, opened 2021-03-16
Update `add` command implementation for implicit this part


darth-cheney:
Right now the grammatical/semantic implementation of the primitive `add/to` command requires that we supply an explicit ObjectSpecifier. We should change this so that leaving off the ObjectSpecifier means, implicitly, to attempt to add the part to the current context part (the one whose script is executing).
  
This came up from @dkrasner 's example here:
https://github.com/UnitedLexCorp/SimpleTalk/pull/133#issuecomment-797729984

dkrasner:
this looks resolved to me @darth-cheney 

i tried some of the examples which were failing earlier, but they no longer are. Perhaps the various merges and conflict resolution took care of it   closing

PR #142, opened 2021-03-16
Implicit Context in Add Command (Issue 139)

darth-cheney:
## What ##
This PR implements a fix for Issue #139, which relates to how the `add` command gets used.
  
#### Summary of Issue ####
Previously, the `add` command always required an explicit ObjectSpecifier as a target, ie:
```simpletalk
add button to second area of this card
```
However, we noticed that this made `tell` versions of this command quite verbose, for example:
```simpletalk
tell third card of current stack to add button to third card of current stack
```
  
This PR represents a grammatical and semantic implementation of the `add` command that, when no ObjectSpecifier is present, will assume the current Part executing the command to be the owner of the new part that gets created. Thus we can now write the above as:
```simpletalk
tell third card of current stack to add button
```
  
As well as, in this example of the script for a Card:
```simpletalk
-- without a button name
add button
-- with a button name
add button "hello"
```
  
## Implementation ##
The implementation adds an additional Command subrule, splitting to two types into `Command_addModel` (which is the form without the ObjectSpecifier) and `Command_addModelTo` (which is the form with a required ObjectSpecifier).
  
We have updated the semantics accordingly, which was a simple change.
  
## Tests ##
I have separated out and updated the [grammar tests](https://github.com/UnitedLexCorp/SimpleTalk/blob/issue-139/js/ohm/tests/test-add-command-grammar.js) for this command, as well as added [specific preload based tests](https://github.com/UnitedLexCorp/SimpleTalk/blob/issue-139/js/objects/tests/test-add-command-with-preload.js) for the cases we care about.
  
## Closes ##
#139


dkrasner:
@darth-cheney this makes sense but something is not working here.

Tests are passing but the specification is failing

image


most examples i tried failed in a similar way

darth-cheney:
@dkrasner I am going to rebase this atop your Target changes once they are in. I think the work we did with non-terminal specifiers on that branch might end up resolving the issues you've highlighted here

dkrasner:
@darth-cheney that makes sense. I think we are almost there with #149

PR #143, opened 2021-03-18
Stop audio's "stop" from re-downloading data

ApproximateIdentity:
This is a pretty small change to make it so that stopping the audio doesn't re-download the beginning of a song. I made it easy to test by making a new example page with a bunch of buttons ("Add audio", "Set name", ...) that sets things up so all you need to do is to click the buttons in the downwards order to test things.

Also I added the example page `js/objects/examples/audio-testing.html` in the first commit and made my code changes in the next commit. That means if you open the page at the current commit (i.e. pointed to by this branch `thomas-audio-stop`), it will run things with the new code. If you instead go one commit back to `thomas-audio-stop~1`, you can use the same example page with the old functionality.

If things are working right, their should be basically no UX changes, but if you open the network tab you will see the old version re-downloads data every time you press stop, but the new version does not.

Let me know what you think.


dkrasner:
not sure how i missed that, yeah this is cleaner (tested works fine)

PR #147, opened 2021-03-19
Fixing issue where Halo resizing was adding px to width prop

darth-cheney:
## What ##
We found that the halo was setting width/height values to be
strings (with px at the end) when users resized using the Halo. This
screws up our ability to use arithmetic in scripting when doing things
with widths and heights. Furthermore, the styler already adds the px
as needed behind the scenes when creating the CSS needed to actually
display the correct width/height values on elements.

I have update accordingly.

Additionally, I have added a new Progress Bar example card to the
layout examples stack snapshot


dkrasner:
@darth-cheney the changes here only concern the layouts-example.html file. Is that correct? or did you forget to commit something

darth-cheney:
@dkrasner ack, you're right. I've pushed the real change now.

PR #149, opened 2021-03-20
Target

dkrasner:
### Main Points ###

This PR adds a "target" property to the Part class, related grammar and semantic handling as well as some tests. 

The target property can be specified as a string if set directly

```
set "target" to "first button of current card"
```

(as before). Or if using the private `setTargetTo` Part command you can also use an ObjectSpecifier. For example,
 
```
setTargetTo "first button of current card"
setTargetTo first button of current card
```
Will effectively reference the same object, the difference being that the grammar-semantics will render the latter `first button of current card` ObjectSpecifier into a partID. 

For this reason "target" is a DynamicProperty which if given a raw ID will convert it to `part id [ID]` in the setter. This ensures that the target value is always a valid ObjectSpecifier. 

Note, that in order for a command to take an ObjectSpecifier argument I have [added it](https://github.com/UnitedLexCorp/SimpleTalk/compare/daniel-target?expand=1#diff-f189d2a29e6e2a0f4a8cd3a2f8c72779e92347a335215074c196e9bec06c5a80R72) to expressions. In general I think this is good and we want to allow handler to take  object as their arguments. 

For example
```
on changeBgColor objSpec
   set "background-color" to "blue" objSpec
end changheColor
```

The reason that `setTargetTo` is a private Part command/handler as opposed to a grammar defined command is because I think that given the choice to make the grammar smaller or the private/public handler count smaller best is to choose the former. The more we can do with said grammar the more powerful it is and adding things to it when absolutely necessary seems like the best way to go. As opposed to commands, private or public, which we can assume will change more. 

One thing that will not work here, and you can see in the example below 

```
set "target" to "button 'Target' of current card"
```
Since we only have single quote literals in the grammar at the moment and this would require a mix of quote-types. I didn't do anything about this at the moment since it's a design decision for the language which I think require more thought. 

here is an example you can run:

image

(which I will upload to slack with the PR)

### DIFF
--- a/js/objects/parts/Part.js
+++ b/js/objects/parts/Part.js
@@ -172,6 +172,10 @@ class Part {
         // to ALL Parts in the system.
         let basicProps = [
             new BasicProperty(
+                'target',
+                null,
+            ),
+            new BasicProperty(
                 'contents',
                 null,
             ),
--- a/js/objects/parts/Part.js
+++ b/js/objects/parts/Part.js
@@ -4,7 +4,10 @@
  * I represent the prototype object for all
  * SimpleTalk parts.
  */
-import idMaker from '../utils/idMaker.js';
+import {
+    idMaker,
+    isValidId
+} from '../utils/id.js';
 import errorHandler from '../utils/errorHandler.js';
 import {
     PartProperties,
@@ -13,6 +16,7 @@ import {
 } from '../properties/PartProperties.js';
 
 import {ActivationContext} from '../ExecutionStack.js';
+
 
 class Part {
     constructor(anOwnerPart, name, deserializing=false){
@@ -236,15 +240,14 @@ class Part {
             'target',
             function(propOwner, propObject, val){
                 // check to see if the target is a non-ID
-                if(val == null){
-                    propObject._value = null;
-                } else if(val.match(/(?!\d)/g).length > 1){
-                    propObject._value = val;
-                } else {
+                let id = isValidId(val);
+                if(id){
                     // if it is an ID insert "part" since our
                     // grammar doesn't handle ID without system object
                     // prefixes
                     propObject._value = `part id ${val}`;
+                } else {
+                    propObject._value = val;
                 }
             },
             function(propOwner, propObject){
--- a/js/objects/parts/Part.js
+++ b/js/objects/parts/Part.js
@@ -66,6 +66,7 @@ class Part {
         this.getOwnerBranch = this.getOwnerBranch.bind(this);
         this.startStepping = this.startStepping.bind(this);
         this.stopStepping = this.stopStepping.bind(this);
+        this.setTargetProp = this.setTargetProp.bind(this);
 
 
         // Finally, we finish initialization
@@ -76,6 +77,7 @@ class Part {
         this.setPrivateCommandHandler("newModel", this.newModelCmdHandler);
         this.setPrivateCommandHandler("openEditor", this.openEditorCmdHandler);
         this.setPrivateCommandHandler("closeEditor", this.closeEditorCmdHandler);
+        this.setPrivateCommandHandler("setTargetTo", this.setTargetProp);
     }
 
     // Convenience getter to get the id
@@ -172,10 +174,6 @@ class Part {
         // to ALL Parts in the system.
         let basicProps = [
             new BasicProperty(
-                'target',
-                null,
-            ),
-            new BasicProperty(
                 'contents',
                 null,
             ),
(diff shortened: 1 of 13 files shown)


js/objects/parts/Part.js
```diff
@@ -230,6 +232,27 @@ class Part {
             [] // No aliases
         );
 
+        this.partProperties.newDynamicProp(
+            'target',
+            function(propOwner, propObject, val){
+                // check to see if the target is a non-ID
+                if(val == null){
+                    propObject._value = null;
+                } else if(val.match(/(?!\d)/g).length > 1){
```
darth-cheney, line 241: Since we want to leave open the possibility of completely changing IDs in the future (for example, to URIs like we were talking about), maybe we can break this out into a util function called like `isValidId()`? That way we don't have to hunt down specific cases like this. In fact, maybe we should have a single ID utils file for now...

darth-cheney (approved):

darth-cheney:
@dkrasner I have made one comment inline, but here are a couple of others:
  
> Since we only have single quote literals in the grammar at the moment and this would require a mix of quote-types. I didn't do anything about this at the moment since it's a design decision for the language which I think require more thought.
  
The original HyperTalk dealt with this using the special keywords `return` and `quote`. For example, your string could be composed using the following interpolation:
```simpletalk
put quote & "Target Name" & quote into targetString
set "target" to "button " & targetString & " of current card
```
  
Also, should we add grammar & preload test cases for referencing the target of a target, or even the target of a target of a target? Or is that too arcane for now?
  
There is an audio related test failing in the full suite run:
```
  Audio Part & View Tests
    System initialialization
      ✓ Can add audio model+view to current card (via message)
      ✓ Can tell audio to loadFromSource (via message)
      ✓ Can tell audio to play (via message)
      ✓ Can tell audio to pause (via message)
      1) Can tell audio to stop i.e. load (via message)


  4 passing (17ms)
  1 failing

  1) Audio Part & View Tests
       System initialialization
         Can tell audio to stop i.e. load (via message):
     AssertionError: expected null to be true
      at Context.<anonymous> (js/objects/tests/test-audio-with-preload.js:85:20)
      at processImmediate (node:internal/timers:464:21)
```

js/objects/parts/Part.js
dkrasner: makes sense. This is too hacky anyhow..

dkrasner:
> @dkrasner I have made one comment inline, but here are a couple of others:
> 
> > Since we only have single quote literals in the grammar at the moment and this would require a mix of quote-types. I didn't do anything about this at the moment since it's a design decision for the language which I think require more thought.
> 
> The original HyperTalk dealt with this using the special keywords `return` and `quote`. For example, your string could be composed using the following interpolation:
> 
> ```
> put quote & "Target Name" & quote into targetString
> set "target" to "button " & targetString & " of current card
> ```
> 
I think people are more used to either using single quotes or escape, like '\' 

Not sure what is best to introduce or whether it should be here. 

> Also, should we add grammar & preload test cases for referencing the target of a target, or even the target of a target of a target? Or is that too arcane for now?
> 
In theory that should be doable
> There is an audio related test failing in the full suite run:
> 
> ```
>   Audio Part & View Tests
>     System initialialization
>       ✓ Can add audio model+view to current card (via message)
>       ✓ Can tell audio to loadFromSource (via message)
>       ✓ Can tell audio to play (via message)
>       ✓ Can tell audio to pause (via message)
>       1) Can tell audio to stop i.e. load (via message)
> 
> 
>   4 passing (17ms)
>   1 failing
> 
>   1) Audio Part & View Tests
>        System initialialization
>          Can tell audio to stop i.e. load (via message):
>      AssertionError: expected null to be true
>       at Context.<anonymous> (js/objects/tests/test-audio-with-preload.js:85:20)
>       at processImmediate (node:internal/timers:464:21)
> ```

coming from PR #143 - will fix it up here

dkrasner:
@darth-cheney 

[this](https://github.com/UnitedLexCorp/SimpleTalk/pull/149/commits/08a2b9424497606bb68fc86151cba4b0efcf1978) takes care of the leftover audio test issues, and cleans up audio

[this](https://github.com/UnitedLexCorp/SimpleTalk/pull/149/commits/2d7d3b418f7220c50c9dc46f85b5966207b131cb) adds isValidId util, makes a single id utility file and adds tests

[this](https://github.com/UnitedLexCorp/SimpleTalk/pull/149/commits/c2ae4a29f37d1cf35057c3f689ba72fd5e40280c) adds further `target of target...` grammar and preload tests. They are clearly not comprehensive but a step in the right direction. The last test in preload has a circular reference (button1 -> button2 -> button3 -> button1) which is kind of fun.

if you think this is good, I'll merge in

darth-cheney:
@dkrasner I just pulled down and tested. It all looks good to me, let's merge it in

Issue #135, opened 2021-03-10
Copy and Paste Parts


darth-cheney:
I could really have used a copy and paste functionality in the course of creating the layout examples stack. With the new serialization pattern it shouldn't be too bad to implement.

dkrasner:
some of this worked in the drag&drop functionality (recall from the world catalogue, really anywhere, you could drag&drop and it would make a copy or a new part), but i think it needs reworking 

one thing to do decide is what a copy is: 
* a new model + view of the same type
* a copy of the model + view, i.e. a new model with all the same part props as the old one
* or a an additional view on the _same_ model

my personal feeling is that it should be the 2nd option, but maybe we need a way to choose between the first two or all three

darth-cheney:
> a copy of the model + view, i.e. a new model with all the same part props as the old one
  
I think this is the right option, because it's the equivalent of a de-serialized version of the original part with a new id, which is exactly what we want. We could simply reuse a lot of that functionality.
  
The ability to copy should be right in the halo (in addition to being a clipboard manipulation, which we should also include). Might be time to start thinking about a custom context menu as well

dkrasner:
yeah maybe all parts could get a "copy" & "paste" commands?

"copy" -> copy the serialization for (selected) object to clipboard
"paste"-> create a new model from the serialization and add it to (selected) object

this way you could script `tell button "My Button" of current card to copy \n tell first window of current card to paste"?

just have to think through this and general clipboard interaction

PR #151, opened 2021-03-23
Copy and Paste (STClipboard v1)

darth-cheney:
## What ##
This PR implements a first pass at Part copy and paste functionality. To accomplish this we introduce a new `STClipboard` object on the System that handles most of the complexity. Note that it is completely separate from the OS/Browser based clipboard and copy/paste actions (see below).
  
### The `STClipboard` Object ###
`STClipboard` represents a kind of SimpleTalk-only clipboard that can be invoked directly by the System. There are two primary methods intended to be the main API:
* `copyPart(aPartObject)` -- This serializes the passed Part into a special store, along with a custom mimeType of `simpletalk/json'`. Note that these serializations, unlike the saveHTML ones, _not_ flattened.
* `pasteContentsInto(aPartObject)` -- This method will attempt to deserialize an object from the clipboard and, if available and if the target part accepts the type, paste a copy of the part into the target object.
  
### A Note on the browser Clipboard API ###
Ideally, we would make extensive use of the [Clipboard API](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard_API), which, with permissions granted, would allow us to intercept the copy and paste actions of the operating system and case based upon the mimeType of true clipboard contents. However, Firefox does not yet support even a functional minimal set of the behavior of this API spec.
  
The objects `STClipboard` and `STClipboardItem` have been designed to be future looking, in that they behave almost like the API will expect down the road. Once that is actually implemented in FF and major browsers, we should be able to make minor changes to have true clipboard functionality without a lot of hassle.
  
#### Copy and Paste Buttons ####
This PR introduces two new buttons for Halos representing copy and paste-into. For the moment these are the only ways to non-JS/console copy and paste Parts.
  
Screenshot from 2021-03-23 14-41-44
  
For the moment the paste-into button is not particularly "smart" -- it will appear on all Parts, regardless of whether or not the STClipboard is empty or, if full, whether or not the target accepts the kind of part that will be pasted. This is something we can deal with using messages and/or events, but we need to make some decisions about how that will work.
  
#### Halo Behavior: Only One & Bug Fixes ####
This PR also fixes an issue where middle click was working for the hierarchical Halo toggling, but shift-click wasn't. Now both should have the same behavior.
  
Additionally, we've made an important change: now only one Halo can be open in the current viewport at a given time. This means that selecting open halo on a part will close all other open halos.
  
## Further Work and Decisions ##
Here are some open questions:
* Do we need to start thinking about contextual menus to make the copy/paste UI a bit cleaner?
* How should we handle the copying/pasting of Cards or whole Stacks (since there is no Halo)?
* We still have a `copyModel` command, but it is kind of dead in the water. We should think about a replacement that makes use of STClipboard to do the heavy lifting. How should these commands be structured?
* Do we want an ST-level message to be sent to all accepting Parts in the current card whenever a new STClipboardItem is added? What should the default behavior be?
  
Also feel free to note any other suggestions you have about what we can do in the future.
  
## Tests ##
There is a [preload test file](https://github.com/UnitedLexCorp/SimpleTalk/blob/eric-copy-paste/js/objects/tests/test-clipboard-with-preload.js) for these changes.
  
## Closes ##
#135

### DIFF
--- a/js/objects/System.js
+++ b/js/objects/System.js
@@ -39,6 +39,7 @@ import interpreterSemantics from '../ohm/interpreter-semantics.js';
 import {ExecutionStack, ActivationContext} from './ExecutionStack.js';
 
 import idMaker from './utils/id.js';
+import STClipboard from './utils/clipboard.js';
 
 import handInterface from './utils/handInterface.js';
 
@@ -88,6 +89,9 @@ const System = {
         } else {
             this.loadFromEmpty();
         }
+
+        // Attach a new clipboard instance
+        this.clipboard = new STClipboard(this);
 
         // By this point we should have a WorldView with
         // a model attached.
--- a/js/objects/parts/Part.js
+++ b/js/objects/parts/Part.js
@@ -42,7 +42,6 @@ class Part {
         this.isPart = true;
 
         // Bind methods
-        this.copy = this.copy.bind(this);
         this.setupProperties = this.setupProperties.bind(this);
         this.setupStyleProperties = this.setupStyleProperties.bind(this);
 
@@ -157,40 +156,6 @@ class Part {
         }
     }
 
-    // perform a deep copy of myself (and all my suparts)
-    // assigning new ids
-    copy(ownerPart){
-        let modelClass = window.System.availableParts[this.type];
-        let model = new modelClass(ownerPart);
-        // cache the model id so it does not get overwritten by
-        // copying partProperties
-        let modelId = model.id;
-        // TODO: we cannot just copy of over neither property subscribers
-        // nor handlers, since these will be incorrectly bound. There is no
-        // native way to clone a js function, although we could implement a
-        // strategy to do so. At the moment any subscribers and handlers that
-        // are not part of the model class natively will be missing from the copy.
-        this.partProperties._properties.forEach((prop) => {
-            model.partProperties.setPropertyNamed(model, prop.name, prop._value);
-        });
-        model.id = modelId;
-        // add the model to the system parts
-        window.System.partsById[model.id] = model;
-        // if there is a script attached to the part we need to compile it
-        let script = model.partProperties.getPropertyNamed(model, "script");
-        if(script){
-            this.sendMessage({
-                type: 'compile',
-                codeString: script,
-                targetId: modelId
-            }, window.System);
-        }
-        // recursively copy each subpart, setting model as its owner
-        model.subparts.forEach((subpart) => {
-            subpart.copy(model);
-        });
-        return model;
-    }
 
     // Configures the specific properties that the
     // given part can expect, along with any default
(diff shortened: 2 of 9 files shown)


ApproximateIdentity:
I'm a little confused what the functionality is here. If I create two buttons and I click copy on one and then paste on the other, what is supposed to happen? Is it supposed to copy the name? The script? Everything? Right now it seems to me like nothing is happening.

darth-cheney:
@ApproximateIdentity Nothing happens because buttons don't accept any subparts (you can't paste a button into another button). Try making a new Area, then pasting the button into that.
  
Right now Cards don't have Halos so we need to think about how this might work for just pasting onto a card.
  
If you want to test this out manually in the browser console, use the two methods mentioned in the PR:
* `System.clipboard.copyPart(aPart)`
* `System.clipboard.pasteContentsInto(aPart)`

Both will take Part (model) objects as arguments. If the target of a pasteInto call accepts the kind of subpart that is in the STClipboard, is should create a copy and put it inside. Otherwise nothing happens.

darth-cheney:
@ApproximateIdentity @dkrasner I've rebased and also added some new functionality, as follows:
* the `paste` command will attempt to paste the contents of the STClipboard into the part that sends it, ie:
  `tell first area of this card to paste`
* the `copy` command will similarly copy the part that sends it into the STClipboard:
   `tell first button of current card to copy`
  
I have decided not to do a contextual menu just yet -- we are going to want the field integration and a couple of other things before we can do it properly. I'm thinking what we have here is enough to get started. Thoughts?

dkrasner:
@darth-cheney this is great

I added a button "Paste here" with the following script
```
on click
   tell current card to paste
end click
```

and then went around copying and pasting into the current card whatever i wanted. 

I imagine that we can combine this with the `onOpenX` command and create stacks/cards by copying from catalogues

```
on openCard
     tell window "Toolbox" of stack "Catalogue" to copy
     tell current card to paste
     tell window "Toolbox" of current card to set "top" to 100
end openCard
```

If we sort out how to deal with implicit 'it' we could write things like `tell it to set...` making pretty readable, flexible scripts. 

there is one test failing
image

which looks like some minor props issue. Otherwise merge it in!

> * Do we need to start thinking about contextual menus to make the copy/paste UI a bit cleaner?
yes but lets see if we can build them out of ST parts eventually
> 
> * How should we handle the copying/pasting of Cards or whole Stacks (since there is no Halo)?
one way to do it is with a custom button like the one i made above. Another way is in your navigator work. Lets see how that pans out and feels. 
At least for the moment it's really easy to make a button on a stack to paste into the current, or any, card.
> 
> * We still have a `copyModel` command, but it is kind of dead in the water. We should think about a replacement that makes use of STClipboard to do the heavy lifting. How should these commands be structured?
> 
i think we should get rid of this and see where copying takes us. 
> * Do we want an ST-level message to be sent to all accepting Parts in the current card whenever a new STClipboardItem is added? What should the default behavior be?
yes I think this is a good idea. Maybe we can have a (private) part prop like "canAcceptClipboard" which would be set on this `newCliopboardItem`   That way halo can simply check if the current part accepts what's in clipboard and shows/hides the paste icon

PR #153, opened 2021-03-26
field refactor1

dkrasner:
### Main Points ###

Like the title above, this PR represents an initial refactor of `st-field`. I imagine there will be one or two more reworkings along these lines.

Outside of the selection related code (more below) `Field` and `FieldView` are much simper than before, and all the editor functionality have been moved out to a separate [EditorView](https://github.com/UnitedLexCorp/SimpleTalk/blob/daniel-field-refactor1/js/objects/views/editors/EditorView.js). 

`st-field` is now standalone element, with a separate editor like `st-button` which I think should be the case for all our part views. It's a slightly stripped down version of the original editor bar, but we can add to it as needed and moreover the editor should really be a SimpleTalk built object. The editor mainly consists of icons which send `propertyChanged` messages to the target model. This should be reasonably straightforward to translate into img-part/views and corresponding click handlers for those. 

`st-field` no longer depends on the `document.execCommand()` browser API, which is deprecated and has unpredictable support. This mainly impacts the select text and edit functionality which has been completely refactored. I consider this to be **experimental** - see more below. 

`st-field` has a `lock` text property which you can toggle in the halo. This is a mode which I have mixed feelings about, but I think it will be useful especially with text selection in the future. 

#### properties ####

Both `text` and `innerHTML` properties have changed for `st-field`: [`text`](https://github.com/UnitedLexCorp/SimpleTalk/blob/daniel-field-refactor1/js/objects/parts/Field.js#L57) is now a DynamicProp which updates the `innerHTML` BasicProperty if `notify=true`. This is coupled with how FieldView [subscribers](https://github.com/UnitedLexCorp/SimpleTalk/blob/daniel-field-refactor1/js/objects/views/FieldView.js#L125) handle the changes. 

Additionally, DynamicProp's now accept a `notify` argument, which you can also see in use in the [`FieldVIew.onInput`](https://github.com/UnitedLexCorp/SimpleTalk/blob/daniel-field-refactor1/js/objects/views/FieldView.js#L237) handler.

The main reason for this sort of rigmarole is that we want to be able to keep the browser context of a contenteditable (cursor, focus etc) but still keep our part properties up to date. The nice thing is that you can now, safely, set `text` and `innerHTML` properties programmatically. 

NOTE: an exceptionally annoying caveat is that JSDOM **does not** automatically updated `element.innerText` when `element.innerHTML` is updated, which is how browsers work and, in part, on what the above part prop dynamic depends. So it makes for hacky tests and some really really fun debugging. 

#### editors ####

There is only one [editor](https://github.com/UnitedLexCorp/SimpleTalk/blob/daniel-field-refactor1/js/objects/views/editors/EditorView.js) now. I couldn't really justify having both a separate field and button editor, b/c they are so close at the moment. And regardless should be written in ST in the future. 

#### Text selection ####

At the moment the only way to edit selected text (for example, to make a chunk bold) is done by pressing the `meta` key and selecting. This will open a new `st-field` populated with the selected range. Editing the new field will edit the corresponding range. Holding down the `meta` key and making multiple selection will open multiple fields. 

Selecting various chunks of text and having those blow up into new fields which you can edit independently, and which merge in seamlessly into the original text is pretty cool but it has issues. 

* the way that the browser `Range` algorithm deals with html merging can sometime cause a new line to appear when first editing. This is annoying.
* if you want to just change the style or color of a selection you don't want a whole new field to open with its own editor. You just want to have the current field editor style the selection. One way to do this is to open a new field with `hide: true` property and then temporarily set the target of the current editor to that field. This would require some thought but in general we should be using the target property here for all editors
* if you make a selection and then edit the original field it will screw up the ranges! maybe we want to lock the original field and unlock when all the selections are closed... not sure
* if you close the original field, the selection field(s) will cause all sorts of target not found errors. Maybe they should all be closed. But in general we need to deal with these types of errors so they don't break the system

#### Tests ####
I have updated all tests, and written new ones for field. A few I have decided to skip, like those for the error handler, due to the  JSDOM idiocy mentioned above.


darth-cheney:
@dkrasner Just playing around here, and two things I've noticed off the bat:
1. There is no longer a way to select the font and font size in the editor tools (we'll need this for example stacks)
2. All the text is gone from the layout example stacks and I'm not sure why

dkrasner:
@darth-cheney the ^ comments are taken care of 

going to merge this in but keep the branch around for a bit just in case

Issue #159, opened 2021-03-30
Drawing and serialization/deserialization


dkrasner:
### Main Points ###

Drawing seems to break on serialization/deserialization. 

The editor tool no longer comes up when clicking the tool-widget and you can get stuck with border like this:

![image](https://user-images.githubusercontent.com/1154326/113012034-b09c4100-917a-11eb-8075-c530973819a7.png)

PR #160, opened 2021-03-31
Postmerge Fixes

darth-cheney:
## What ##
This PR represents a few bugfixes for issues we discovered after the most recent merge. They include:
* Changes to how `getCurrentCardModel` etc are implemented and used (this affects some aspects of navigation);
* Ensuring that `add <partName>` without an object specifier can deal with ambiguity appropriately (using message delegation)
* Fixing the issue identified in Issue #159 
  
There are still some stack navigation tests failing for reasons I have not yet been able to determine, but that situation already exists on master.
  
## Closes ##
#159

### DIFF
--- a/js/objects/System.js
+++ b/js/objects/System.js
@@ -701,33 +701,19 @@ const System = {
 
     // return the model corresponding to the current stack
     getCurrentStackModel: function(){
-        // make sure there is only one current-stack
-        let currentStacks = document.querySelectorAll('st-world > st-stack.current-stack');
-        if(currentStacks.length > 1){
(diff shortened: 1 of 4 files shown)


dkrasner:
@darth-cheney this all looks good and I am happy to merge it in

there is still a stack navigation test failing (only one left which was failing in `master`)  do you want to add a fix to that here as well?

darth-cheney:
@dkrasner No let's merge these in first. We need to come back to that issue. It's already present on master but it's such a mind bender that it might as well be its own issue

PR #164, opened 2021-04-01
fixing "doIt" related functionality

dkrasner:
### Main Points ###

doIt button dissapears with subsequent clicks
selection text is cleaned up, newlines/spaces removed
MNU error handler checks for error message context and inserts markup accordingly

error tests pass


dkrasner:
additionally:

GrammarMatchError is handled/marked up in the field on a 'doIt' message error
private command handlers are part of the registry with a key `private: true`

darth-cheney:
Looks good to me

Issue #129, opened 2021-03-03
Global interruput for all stepping


dkrasner:
### Main points ###
 

you can easily create runaway processes which make the browser unresponsive, or make debugging impossible (for example an MNU error in the `step` handler can bring up error fields preventing any interaction). 

I propose:
* global interrup keyboard shortcut like `ctrl-C` which will send a `set "stepping" to false` in all models, this can be stupid at first just sending this message to all button, field, image, container, window and drawing part model (everything that has a `stepping` property
* in the case of an error within a `step` handler a `set "stepping" to false` message should first be sent to the target and then the error handled

PR #165, opened 2021-04-02
adding a global Interrupt

dkrasner:
### Main points ###

This PR adds a `globalInterrupt` command handler to the System. At the moment it cycles through all parts and set `stepping` to false in those which are stepping. The handler can be called either via message or by keydown ctrl-c combo. 

Tests were added. 

This PR closes issue #129

### DIFF
--- a/js/objects/System.js
+++ b/js/objects/System.js
@@ -1450,6 +1450,15 @@ System._commandHandlers['toggleHandDetection'] = () => {
     }
 };
 
+System._commandHandlers['globalInterrupt'] = () => {
+    // cycle through all the parts and set the "stepping" property to false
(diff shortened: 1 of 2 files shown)


js/objects/System.js
```diff
@@ -1523,6 +1532,18 @@ document.addEventListener('DOMContentLoaded', () => {
     System.initialLoad();
 });
 
+// global interrupt
+document.addEventListener('keydown', (event) => {
```
darth-cheney, line 1536: Good for now. We'll come back to this once we deal with keybindings in the System etc.

dkrasner: exactly

dkrasner:
@darth-cheney merge?

PR #166, opened 2021-04-05
Navigator v1

darth-cheney:
## What ##
This PR implements a new basic System accessible webcomponent called `Navigator`. It provides an interactive view of all available stacks and cards on those stacks. 
Screenshot from 2021-04-05 15-52-42
  
### "Lensing" ###
We introduce the concept of "lensing," whereby a given PartView subclass has `part-id` replaced with `lens-part-id` but maintains the same id mapping. This allows us to recreate a tree of subviews that will be the same as the main Part they are "mapping," while not being the primary views. This also allows the "lensed" views to be bound to and react to the same model as their primaries.
  
In short, this is how the miniature Card and Stack views work inside the Navigator -- they contain nested "lensed" views that are still able to react to changes on the common model all the way down.
  
### Navigator Webcomponents ###
Navigator is not a new kind of Part, but it does have components that are themselves new kinds of PartViews. This allows certain pieces of the Navigator to react to model changes. For example, the Navigator itself is a PartView that expects to take the WorldStack as its model.
  
Here is a description of the new components that can be found in `views/navigator/`:
| Component Name | PartView Subclass? | Description |
|--------------------------|----------------------------|-----------------|
| **`Navigator`** | Y | This is the main component. At initialization, it is attached to the bottom of the webpage. It can be accessed via the DOM (as `st-navigator`) or directly from `System.navigator`. It takes the `WorldStack` as its model. |
| **`CardRow`** | Y | This view represents a row of miniaturized, clickable views of a given card. Its actual child is a lensed view copy of the real card, and its model is the given Card model. It utilizes a `WrappedView` when presenting each mini card. |
| **`StackRow`** | Y | This view represents a row of miniaturized, clickable views of a given stack. Its actual child is a lensed view copy of the real stack, and its model is the given Stack model. It utilizes `WrappedView` when presenting each mini stack. |
| **`WrappedView`** | N | This web component is not a PartView, but rather a utility that handles measuring, scaling, and presenting the miniaturized version of the lensed stack or card. |
  
### Navigation and Reaction ###
Because the `Navigator` and `StackRow`/`CardRow` are all PartViews with the relevant views attached, normal navigation messages in the wider system will result in reactive changes in the navigator. Additionally, clicking on a wrapped Stack in the StackRow will navigate to that Stack in the live environment. It will also re-load the CardRow beneath it because the `current-stack` model is now different. The CardRow views behave similarly -- clicking a card will navigate to it.
  
Note also that because the various `WrappedViews` contain lensed copies of view trees that are themselves bound to real models, changes to parts on a given Card or Stack should be visible, live, in the miniature navigator views.
  
## Usage ##
For now there are two ways to toggle the Navigator:
1. Directly call `System.navigator.toggle()` in the browser console;
2. Use the set key combination of Ctrl + Alt + Spacebar to toggle
  
## Discussion ##
Let's use this PR to discuss any additions or changes we would like to make to the Navigator. We should also test it here thoroughly as we can and decide what works best.  

## Tests ##
Right now I have no tests at all for these, as I'm not sure how exactly I would do it given the messages flying around and all of the timing issues.

### DIFF
--- a/css/core.css
+++ b/css/core.css
@@ -45,6 +45,7 @@ body {
     height: 100vh;
     overflow: hidden;
     font-family: "Crimson Pro";
+    background-color: purple;
 }
 
 .active {
@@ -309,7 +310,8 @@ wrapped-view st-stack st-card:not(.current-card){
     display: none;
 }
 
-wrapped-view > st-card {
+wrapped-view > st-card,
+wrapped-view > st-stack {
     display: block;
 }
 
@@ -328,4 +330,11 @@ wrapped-view > st-card {
 
 .editable {
     outline: solid 1px var(--palette-orange);
+}
+
+
+/* = TESTING: TODO: Remove
+ * --------------------------------------------------*/
+st-navigator {
+    background-color: rgba(255, 255, 255, 0.2);
 }
--- a/css/core.css
+++ b/css/core.css
@@ -335,6 +335,3 @@ wrapped-view > st-stack {
 
 /* = TESTING: TODO: Remove
  * --------------------------------------------------*/
-st-navigator {
-    background-color: rgba(255, 255, 255, 0.2);
-}
--- a/css/core.css
+++ b/css/core.css
@@ -83,7 +83,8 @@ st-card {
     height: 100%;
 }
 
-.current-card {
+.current-card,
+st-card[role="lens"] {
     display: block;
     visibility: initial;
     background-color: var(--palette-blue);
--- a/css/core.css
+++ b/css/core.css
@@ -297,6 +297,22 @@ st-world > st-stack > :not(.current-card).list-layout {
     left: unset !important;
 }
 
+/* =Layout Exceptions
+ * --------------------------------------------------------*/
+/* We don't want list-layout to display on Cards when they
+ * are not current in the current stack (but we want this
+ * in stack windows) */
+/* st-stack.current-stack > st-card.list-layout:not(.current-card){ */
+/*     display: none; */
+/* } */
+wrapped-view st-stack st-card:not(.current-card){
+    display: none;
+}
+
+wrapped-view > st-card {
+    display: block;
+}
+
 /* =Utility Classes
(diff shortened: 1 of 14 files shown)


dkrasner:
This is great! and i'd like to get this in asap. 

A small issue i found playing around for 5mins:

When i imported `bootstrap.html` into the layout-examples.html I saw this:
image

Somehow the color of the lens-card is white, but all the parts are there, i.e. if i add a part, move it around it's there.

darth-cheney:
> also clicking back and forth between stacks broke then app
  
I get this from time to time also, but can't tell what triggers it. If you can figure out a way to reproduce reliably definitely post it here.
  
> Somehow the color of the lens-card is white, but all the parts are there, i.e. if i add a part, move it around it's there.
  
Yeah it looks like we need to make a decision about styling here. The blue color you see is because of the normal `.current-card` class being given that background color styling in the core CSS file [here](https://github.com/UnitedLexCorp/SimpleTalk/blob/master/css/core.css#L89). The actual `background-color` part style property for that Card appears to be `null`, which is the default. I'm thinking we should give specific defaults to all cards at construction time and just decide what that color should be. We should probably remove that line from the core CSS file since it overrides our properties behavior and makes things a little more confusing...

dkrasner:
> class being given that background color styling in the core CSS file [here](https://github.com/UnitedLexCorp/SimpleTalk/blob/master/css/core.css#L89). The actual `background-color` part style property for that Card appears to be `null`, which is the default. I'm thinking we should give specific defaults to all cards at construction time and just decide what that color should be. We should probably remove that line from the core CSS file since it overrides our properties behavior and makes things a little more confusing...

leave this till i PR - merge in `daniel-style-fixes`

dkrasner:
> > also clicking back and forth between stacks broke then app
> 
> I get this from time to time also, but can't tell what triggers it. If you can figure out a way to reproduce reliably definitely post it here.

i get this clicking back and forth and the error posted seems to be present every time

PR #168, opened 2021-04-06
"is there a" conditional

dkrasner:
### Main Points ###

This PR adds a `is there a [ObjectSpecfier]` conditional to the grammar and semantics. Now you can write fun scripts like

```
on click
	if is there a button "New Button"
	then
		answer "Found New Button"
	else
		add button "New Button"
	end if
end click
```

First click will add "New Button" and subsequent clicks will answer. 

Note: I have called this rule `IsThereAnObjectConditional` thinking we might want to have other `IsThereAn` rules in the future (for example `IsThereAVariable` etc) and it would be easiest to keep these as separate rules.
Also I did not add a `is there not a [ObjectSpecifier]` rule since you can handle it in an `if .. then .. ` statement. But if you like I can easily add it. 

Tests have been written for both grammar and semantics.


darth-cheney:
We def need this. A few comments:
* Don't we want it to be "there is a" rather than "is there a"?
* We should be optional with the "a"/"an", especially since it has no function once parsed [ `("a" | "an")` in the rule ]
* I'm thinking that the grammar rule itself should be an Expression rather than a Command, since all of our other commands result in message sends. I also think you might be able to easily make the argument being evaluated a Factor without any change in behavior (ie, ` "there" "is" ("a" | "an") Factor`)
  
What do you think?

dkrasner:
@darth-cheney 

> We def need this. A few comments:
> 
>     * Don't we want it to be "there is a" rather than "is there a"?
sure can change
> 
>     * We should be optional with the "a"/"an", especially since it has no function once parsed [ `("a" | "an")` in the rule ]
yeah that's also easy
> 
>     * I'm thinking that the grammar rule itself should be an Expression rather than a Command, since all of our other commands result in message sends. I also think you might be able to easily make the argument being evaluated a Factor without any change in behavior (ie, ` "there" "is" ("a" | "an") Factor`)
not sure what you mean. This is **not** a Command it's a Conditional, i.e. something that results in a bool. 
> 
> 
> What do you think?

darth-cheney:
> not sure what you mean. This is not a Command it's a Conditional, i.e. something that results in a bool.
  
Whoops. Was reading the code in the first commit (where it was listed as a command). I see the full thing now.

dkrasner:
> > not sure what you mean. This is not a Command it's a Conditional, i.e. something that results in a bool.
> 
> Whoops. Was reading the code in the first commit (where it was listed as a command). I see the full thing now.
oh yeah i was being a total idiot this am


but i was also thinking of making it more flexible (including expressions potentially) in the way you are saying I think, and then decided I wasn't sure what we would want in the end and it just made sense to wait 
a
`"there is (a | an)" ObjectSpecifier` is very clear in the grammar and semantics and easy to debug. If we want other things, whatever those might be, it might make sense to branch the rules or combine them tbd but do so down the line once we use it some more

what do you think?

darth-cheney:
Yeah just leave it at that for now, we can revisit if we need extra functionality.

dkrasner:
@darth-cheney @ApproximateIdentity 

ok grammar (etc) has been updated to reflect the more natural "there is (a|an)" 

and i decided to add the "there is not (a|an)" 

so now you can write succinct things like:
```
on click
	if there is not a button "New Button"
	then add button "New Button"
end click
```

which sounds pretty good when you read it

squash/merge?

darth-cheney:
yeah merge it in

PR #170, opened 2021-04-07
Serialization v2

darth-cheney:
## What ##
This PR introduces new serialization classes separate from the System. It should  make understanding and debugging the serialization/deserialization process a bit better. It has also improved the performance of operations like `importWorld` dramatically.
  
### New Classes ###
We introduce [two new classes](https://github.com/UnitedLexCorp/SimpleTalk/blob/eric-modular-serialization/js/objects/utils/serialization.js): `STDeserializer` and `STSerializer`. Each take a System as their constructor's sole argument. They are designed to be used not only by System, but by any tool that requires serializing/deseralizing. For example, the clipboard has been updated to exclusively use these classes in copying and pasting Part information.
  
### Async Deserialization ###
`STDeserializer.deserialize()` is **now an asychronous operation**. Here we are using Promises to deal with that. Aside from fetches, this is the first truly async operation we've incorporated into the SimpleTalk system. As such, I have made minor changes elsewhere to deal with this fact -- including in some tests that now need to know to wait for deserialization to complete.
  
`STDeserializer.serialize()`, however, is still synchronous since it is quite simple.
   
### Versioning ###
I am also introducing a new piece of information in each serialization: a version string. This should help us determine serialization format versions as we rapidly make changes to the system, and we might be able to load older versions of the serializer/deserializer as drop in replacements when needed.
  
## Trying It Out ##
Though the automated tests are passing (see below), we will still want to bang away at this ourselves. There are generally three things to try out:
1. Saving whole snapshots and successfully reloading from them;
2. Copying and pasting parts;
3. Using `importWorld` without issue
  
Each of us should try all of these and report any problems in this PR.
  
## Future Work & Additional Details ##
We might consider dropping the Promises in favor of the more modern [async/await syntax](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Asynchronous/Async_await) that is now widely available. If you all agree, I think we should open an issue and pursue it in a subsequent PR.
  
The deserializer is currently quite verbose. This is for debugging purposes, but also for giving us the space to add any "lifecycle hooks" should we feel we need them at any point down the road. The implementation also makes extensive use of private caches. It is unclear what the actual performance improvements of using these caches is since I implemented them all at once, but the overall deserialization performance is much better now even on larger stacks.
  
## Tests ##
Since these classes are a drop in replacement, we didn't add any new tests. Rather, we updated (for async compatibility) and adapted to the existing serialization and related test. Those are all now passing.

### DIFF
--- a/js/objects/utils/serialization.js
+++ b/js/objects/utils/serialization.js
@@ -54,6 +54,7 @@ class STDeserializer {
         this.handleId = this.handleId.bind(this);
         this.throwError = this.throwError.bind(this);
         this.flushCaches = this.flushCaches.bind(this);
+        this.dispatchViewAdded = this.dispatchViewAdded.bind(this);
     }
 
     deserialize(aJSONString){
@@ -88,6 +89,7 @@ class STDeserializer {
                 } else {
                     let targetView = document.querySelector(`[part-id="${this.targetId}"]`);
                     targetView.appendChild(rootView);
+                    this.dispatchViewAdded(rootView);
                 }
                 return this;
             });
@@ -202,6 +204,7 @@ class STDeserializer {
                     let view = this._viewsCache[rootPart.id];
                     target.addPart(rootPart);
                     targetView.appendChild(view);
+                    this.dispatchViewAdded(view);
                 });
             });
     }
@@ -332,6 +335,7 @@ class STDeserializer {
         } else {
             document.body.prepend(this.rootViews[0]);
         }
+        this.dispatchViewAdded(document.querySelector('st-world'));
     }
 
     getFlattenedPartTree(aPart, list=[]){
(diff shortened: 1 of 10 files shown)


js/objects/utils/serialization.js
```diff
@@ -0,0 +1,434 @@
...
+        // ie full deserialization.
+        this.targetId = 'system';
+
+        // By default, we create new IDs for each
+        // deserialized Part (and preserve a mapping
+        // from old to new temporarily).
+        // However, there are cases, like initialLoad,
+        // where we want to preserve the originals.
+        // That flag is set here.
+        this.useOriginalIds = false;
```
dkrasner, line 36: i feel like it's time we roll with a UUID setup in the near future

dkrasner:
@darth-cheney the code looks clean. I don't have much to comment there. 

on the await/async cs promises... current it's very clear to read so maybe this can wait until a more major serialization refactoring

i tried various things and the only issue i encountered was preservation of current card with multiple stacks  but I can't reproduce it every single time so not sure how helpful this is... 

here is the recipe i just followed:

start with bootstrap
import layouts
go to some random card (which was not the first/current one)
save
reload snapshot

I've seen it load to the first card of layouts

again since i can't get this to happen every time we can let it go in my view, see what shakes out

PR #171, opened 2021-04-07
Borders shadows corners

dkrasner:
### Main Points ###

This PR adds borders, shadow and corner rounding to ST Style Properties. 

The goal here was too keep the language as explicit, and simple, as possible but also to allow as much control as possible. This lead to some similarities and difference to what you might expect with css. 

Similarities with css: 
you can control any border (top, bottom, right, left) independently. For example, you can say `set "border-top-style" to "ridge"`
and you can set all the borders at the same time. For example, `set "border-style" to "ridge"`. This will iterate through and set all the underlying css borders (this was cleaner than dealing with css string parsing and combing property values in some css specified order which is hard to remember)

Differences with css:
properties related to css colors are split into two ST properties: `color` and `transparency`. For example, you can `set "border-color" to "red"` and `set "border-transparency" to 0.5`. This in line with our previous uses of color and transparency. 

css shadows allow multiple shadows to be added to an element. We only allow one. There are ways to make the `shadow-*` a list, but I think one shadow for now is ok. And it's clear what's what. 

Here is a full list:
"border-style"
"border-top-style"
"border-bottom-style"
"border-left-style"
"border-right-style"

"border-width"
"border-top-width"
"border-bottom-width"
"border-left-width":
 "border-right-width"

"border-color"
 "border-top-color"
 "border-bottom-color"
"border-top-color"
"border-right-color"

"border-transparency"
"border-top-transparency"
"border-bottom-transparency"
"border-top-transparency"
"border-right-transparency"         

"corner-round":
"corner-top-left-round"
"corner-top-right-round"
"corner-bottom-left-round"
"corner-bottom-right-round"

"shadow-left":
"shadow-top"
"shadow-blur"
"shadow-spread"
"shadow-color"
"shadow-transparency"

#### Tests ####

I have written pretty extensive tests for the styler. 

#### to Note ####

I have some defaults for button and field, just as a start. 
These have not been integrated into the st-editor. If we see that the editor is not going to be built in ST anytime soon, I can add them in.
This PR builds on #169 so we can me merged sequentially or we can ignore #169 and only merge this one. 

#### for fun ####

Now you can do something like this:
```
on click
	set "width" to  300
	set "height" to  300
	set "border-right-style" to "ridge"
	set "border-right-width" to "thick"
	set "border-left-style" to "double"
	set "border-left-width" to "thick"
	set "border-top-style" to "dashed"
	set "border-top-width" to 10
	set "corner-top-left-round" to 20
	set "corner-bottom-left-round" to 40
	set "corner-top-right-round" to 10
	set "corner-bottom-right-round" to 80
	set "shadow-left" to 10
	set "shadow-top" to 20
	set "shadow-color" to "green"
	set "shadow-transparency" to 0.3
	set "shadow-spread" to 10
	set "text-color" to "red"
	set "text-size" to 50
	set "name" to "I am ugggly..."
end click
```
image


darth-cheney:
@dkrasner This (and the previous branch) seem to work pretty well with only one issue that I've noticed:
the border values are not preserved when copying/pasting and when deserializing a saved snapshot. You can test this out by making some new areas and copying/pasting them or even saving a basic bootstrap-derived snapshot. Event the toolbox buttons lose their borders when reloading.
Screenshot from 2021-04-07 15-13-11

dkrasner:
@darth-cheney this is happening only for area right?

I just tried to copy and paste the _ugly button_ and it worked 
image


I also copy and pasted a toolbox button from a freshly loaded (deserialized) bootstrap file: 
image


Note: the properties are all there on the original button (those you see in the console) and they are also present both as props as in the inline style in the pasted button (you see the shadow). The default border width on the toolbox buttons is set to 0 so you are not supposed to see it.

darth-cheney:
@dkrasner No I have this problem with the Toolbox (see in my screenshot, there are no button borders) whenever I save a snapshot and load it again

dkrasner:
ok I see it now...

I wonder if this is an ordering issue, i.e. setting individual sides and all sides is not compatible. Whichever comes last wins. 


hmmm

i could get rid of the setting all sides property for now and we can make that a utility, i.e.

there would be no such things as 
`border-style`

but only

`border-style-top`
`border-style-left`
`border-style-bottom`
`border-style-right`

and all compound operations can be handled with a utility handler written in ST 

`tell [] to setBorderStyle "name"`
or 
`tell [] to setBorder "style" "name"` 


what do you think?

darth-cheney:
> what do you think?
  
You can get around this by having a single dynamic prop for setting all sides, similar to [what I did here for "pinning"](https://github.com/UnitedLexCorp/SimpleTalk/blob/master/js/objects/utils/styleProperties.js#L136). Note how the underlying properties being set are intentionally not sending out notifications of change -- only the dynamic prop does that

dkrasner:
yeah i thought about it.

It would either not have a getter at all, or would only return a value if all the sides agreed...

dkrasner:
@darth-cheney ok i've cleaned this up, removing all composite style props; as discussed these can be added in ST itself either as command handler or as GUI's

the copy& paste issue seems to be solved, here is an example for area:
image

i've updated the bootstrap serialization, since we needed a new version

PR #172, opened 2021-04-09
Thomas merriam api

ApproximateIdentity:
This adds an extremely basic use of the merriam api. Create a button with the following script

```
on click
	merriam
end click

on merriamresult result
	answer result
end merriamresult 
```

and then click on it and you'll see some merriam results coming back.

What more do we want to add to this for the moment?

### DIFF
--- a/js/objects/System.js
+++ b/js/objects/System.js
@@ -1289,7 +1289,8 @@ System._commandHandlers['toggleHandDetection'] = () => {
     }
 };
 
-System._commandHandlers['merriam'] = () => {
+System._commandHandlers['merriam'] = (senders) => {
+    const sender = System.partsById[senders[0].id];
     const url = "https://patents.merriamtech.com/_api/merriam/"
     const payload = {
         "fields": [
@@ -1315,7 +1316,14 @@ System._commandHandlers['merriam'] = () => {
     fetch(url, params).then(data => {
         return data.json();
     }).then(json => {
-        console.log(json);
+        const msg = {
+            type: 'command',
+            commandName: 'merriamresult',
+            args: [
+                JSON.stringify(json)
+            ]
+        };
+        sender.sendMessage(msg, sender);
     });
 };
 
--- a/js/objects/System.js
+++ b/js/objects/System.js
@@ -1289,6 +1289,36 @@ System._commandHandlers['toggleHandDetection'] = () => {
     }
 };
 
+System._commandHandlers['merriam'] = () => {
+    const url = "https://patents.merriamtech.com/_api/merriam/"
+    const payload = {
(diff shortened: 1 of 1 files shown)


dkrasner:
@ApproximateIdentity cool

i would just add an id argument `merriam docId` later we can add the weight floats in some ST like way

and i would move the fetch code out to some importable file (yes small for now but will grow, since we'll want the api to have a generator like capability, as in `merriam docId "next"`_

 <- actually maybe the handler could take take two arguments `docId` and `which (="next" | "previous" or empty)"`

and so the little merriam middleware could be smart:
`merriam doId "next"` will return the next document in the queue (array) of cached response results. If no results are cached it will fetch them first and return the first id in the list)
`merriam docId "previous"` will return the previous document

next/previous could will be modulo the result total (which we could hardcode to 100 or leave the API default)

this way we'd have a flexible interface for working with multiple documents. `merriam docId1 "next"` and `merriam docId2 "next"` will return their respective (cached) results. 

we'll work out some similar id thing for snippets and be golden. 

(for now i would leave all the arguments to the handler as string literals, handled by the middleware under the hood. As opposed to doing things like `merriam docId next` and having to add parsing of this to the grammar/semantics)

thoughts?

ApproximateIdentity:
I pushed changes that allow changing the doc id and moving the file to a separate utility file.

As for the the generator capability I'm not totally sure what's best there. I mean what does that even have to do with the merriam api anyway? Wouldn't any thing iterable be something that could support that sort of thing? Maybe it's better to have this sort of generator like stuff build in deeper and independent of this?

dkrasner:
it doesn't have anything to do with merriam, except that we just a way to get the next result so it has to happen somewhere and this seems like a natural place since this api util would be the one to keep track of requests/responses and so could know what last returned id was, what "next" one would be etc

ApproximateIdentity:
The merriam api allows for passing limit and offset so why don't we just figure out a way to pass those parameters through? Then the user can do that. I just don't feel like the addition of state into the api is a good idea. I don't really know how I would implement it at a high-level (i.e. ignoring js/simple talk/etc.).

dkrasner:
ok we can cross that bridge later

@darth-cheney you good with this?

darth-cheney:
@ApproximateIdentity @dkrasner this looks good to me

ApproximateIdentity:
I think we wanted to wait to merge this until after we worked out a bunch of the bugs since it's so basic. Then again I guess this is so localized that merging it won't really make the bugs easier or harder to squash. Anyway it's not a huge prio.

dkrasner:
@ApproximateIdentity merge whenever

Issue #180, opened 2021-04-12
Navigator and layouts don't play nice


dkrasner:
Navigator can start to push the world/stack/card layout up off-screen like so

![image](https://user-images.githubusercontent.com/1154326/114387386-a9216280-9b92-11eb-8780-4cac40c5ecc1.png)

these can be cumulative. example opening subsequent script editors can push the navigator up

This effectively kills the system since there is no way to access parts pinned at the top

darth-cheney:
Closed by #183

Issue #179, opened 2021-04-12
Navigator gets stuck crashing the system


dkrasner:
hard to reproduce deterministically but this is quite common:

![image](https://user-images.githubusercontent.com/1154326/114387279-85f6b300-9b92-11eb-8503-a1c481b4f290.png)

sometimes it just doesn't open at all 

sometimes it opens but no stacks or cards appear:
![image](https://user-images.githubusercontent.com/1154326/114399127-dffe7500-9ba0-11eb-9751-6e27a569f950.png)

or it appears doubling up
![image](https://user-images.githubusercontent.com/1154326/114399244-fd334380-9ba0-11eb-9953-224785474db3.png)

darth-cheney:
Closed by #183

PR #183, opened 2021-04-12
Navigator Fixes and Revisions

darth-cheney:
## What ##
This PR fixes display and logic bugginess in the Navigator.
  
## Problems ##
There were several cascading problems, partially identified in #179  and #180, that were causing Navigator to completely bug-out and at times freeze the whole system.
  
There were a few things going on. First, flipping back and forth quickly between two Stacks with the Nav open could lead to display bugs. This is because the CardRow for each Stack was being composed and re-rendered on the fly each time, meaning a whole tree of elements was being copied, inserted, and having its model set (recursively) each time we switched between stacks.
  
The "layout pushes up" issue was a fairly arcane one. It only applied to `FieldView` and no other parts. It turns out that `FieldView` sets its element focus when it is attached to the DOM. Because the Navigator's "lensed views" are just PartView elements with the same model attached, the browser was attempting to focus the tiny, scaled-down field in the Nav, meaning it would scroll (without scrollbars) to that element to get it within the viewport. This is why things seemed to "push up"
  
### `WrappedView` is now a PartView ###
Previously, WrappedView was simply a custom webcomponent that had no strict relation to the rest of the SimpleTalk system. But now we've made the better decision to make it a PartView subclass, meaning it can accept a model and has all the lifecycle methods we would expect from a normal PartView.
  
Instead of waiting for some user action or for the right DOM to be set, the nested "lensed" trees will be created whenever the model is set on a WrappedView. This is also when the scaling of the view will occur. Consequently, "currentness" can now apply to the WrappedView as well, where needed.
  
### Optimizing CardRow Rendering ###
As mentioned, we were re-rendering the CardRow WrappedViews each time we switched between Stacks. That's a lot of computation to happen if you are toggling back and forth quickly. The main Navigator PartView will now pre-render all CardRows for any Stack that is loaded into the System, showing only the CardRow that is mapped to the current stack.
  
The latter process is interesting: we simply set the `slot=` attribute on the `<nav-card-row>` (a CardRow instance) that maps to the current stack, and we remove that attribute on all others. This results in an instant swapping, since CardRow must be specifically slotted in Navigator's shadow dom.
  
### The "pushing-up" issue ###
The solution here was crude but effective: remove the line where FieldView automatically gets the focus whenever it is attached to the DOM. We definitely don't want fields inside of the nav trying to get the focus. There might be a better way down the road to ensure that the main FieldViews do get the focus that we want while ensuring that WrappedView copies are not affected. TBD.
  
Toggling should now happen quickly and without bugginess.
  
### Other Fixes ###
I've added some minor styling fixes here and there. For example, Cards that make use of a list-layout on themselves should now display correctly in the Nav WrappedViews.
  
## What To Do ##
Steps to really put the Navigator through the ringer:
1. Load the layout stack and click through cards. Toggle nav with Ctrl+Alt+Space and really try to break it.
2. Load another stack using `importWorld "layout-examples.html"` or whatever stack you want. Note that at this point in the branching any stack with a Drawing in it will not load properly
  
## Tests ##
We had to update some querySelectors in a few tests. These selectors assumed only a single stack/card tree existing in the document. But with Nav we can have copies of stacks and cards inside of the navigator. Some of the tests were pulling out the copies and not the main views to be tested. The solution is to prepend `st-world` to any relevant querySelector, which will then not include anything in the navigator.
  
## Closes ##
#179 
#180

### DIFF
--- a/css/core.css
+++ b/css/core.css
@@ -23,8 +23,8 @@
     --container-default-height: 15vh;
     --drawing-default-width: 500px;
     --drawing-default-height: 300px;
-    --field-default-width: 40vw;
-    --field-default-height: 30vh;
+    --field-default-width: 300px;
+    --field-default-height: 200px;
 }
 
 /* Keyframe Animations */
@@ -41,8 +41,7 @@
 
 body {
     display: block;
-    width: 100vw;
-    height: 100vh;
+    position: relative;
     overflow: hidden;
     font-family: "Crimson Pro";
     background-color: purple;
@@ -55,7 +54,11 @@ body {
 /* =World Styles
  *----------------------------------------------------*/
 st-world {
-    visibility: hidden;
+    display: block;
+    position: absolute;
+    overflow: hidden;
+    width: 100vw;
+    height: 100vh;
 }
 
 
@@ -173,6 +176,7 @@ st-field {
     height: var(--field-default-height);
     background-color: var(--palette-cornsik);
     position: absolute;
+    display: block;
 }
 
 /* =Layout Class Styles
--- a/js/objects/views/FieldView.js
+++ b/js/objects/views/FieldView.js
@@ -150,8 +150,6 @@ class FieldView extends PartView {
 
         this.textarea.addEventListener('input', this.onInput);
         this.textarea.addEventListener('keydown', this.onKeydown);
-        this.textarea.focus();
-        // document.execCommand("defaultParagraphSeparator", false, "br");
         // No need to add a click listener as the base PartView class does that
     }
 
--- a/js/objects/views/navigator/WrappedView.js
+++ b/js/objects/views/navigator/WrappedView.js
@@ -7,6 +7,8 @@
  * I make a cloned copy of the underlying view and
  * attach it to the same model as the original.
  **/
+import PartView from '../PartView.js';
+
 const templateString = `
 <style>
     :host {
@@ -33,14 +35,7 @@ const templateString = `
         transition: box-shadow 0.1s linear, border 0.1s linear;
     }
 
-    :host(.hide),
-    :host(.hide.current){
-        transform: translateY(100%);
-        opacity: 0.01;
-        transition: transform 0.2s ease-out, opacity 0.2s ease-out, box-shadow 0.2s linear;
-    }
-
-    :host(.hide) > st-card {
+    :host(:not(.current)) > st-card {
         display: none;
     }
 
@@ -89,7 +84,7 @@ const templateString = `
 `;
 
 
-class WrappedView extends HTMLElement {
+class WrappedView extends PartView {
     constructor(){
         super();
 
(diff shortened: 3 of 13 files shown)


js/objects/views/FieldView.js
```diff
@@ -150,8 +150,6 @@ class FieldView extends PartView {
 
         this.textarea.addEventListener('input', this.onInput);
         this.textarea.addEventListener('keydown', this.onKeydown);
-        this.textarea.focus();
```
dkrasner, line 153: removing this is fine. It was really there b/c it was natural to open the field and be able to start typing in it right away (as opposed to open the field, click in it, then start typing). 

another option is to have `focus` set when respond to the `editable` property being set to `true` [here](https://github.com/UnitedLexCorp/SimpleTalk/blob/905d6079767c4784bf332fa3a2fd107641800c83/js/objects/views/FieldView.js#L118) 

"lens" elements can have `editable` always set to false

css/core.css
```diff
@@ -23,8 +23,8 @@
     --container-default-height: 15vh;
     --drawing-default-width: 500px;
     --drawing-default-height: 300px;
-    --field-default-width: 40vw;
```
dkrasner, line 26: what's the issue with having viewport relative units here?

js/objects/views/navigator/WrappedView.js
```diff
@@ -155,12 +157,67 @@ class WrappedView extends HTMLElement {
...
+        // Insert the root lensed view into the wrapper
+        this.setAttribute('wrapped-id', aPartModel.id);
+        this.appendChild(lensedView);
+        this.updateScaling();
+    }
+
+    _recursivelyUpdateLensViews(lensedView, anId){
+        let subViews = Array.from(lensedView.children);
+        subViews.forEach(subView => {
+            subView.isLensed = true;
```
dkrasner, line 199: this could be a place to set the `editable` property to `false` if the subView is a field.

dkrasner:
@darth-cheney this looks good and it feels a lot snappier. Some comments made above. I wasn't able to break it (yet) within an environment but importing a few worlds then serializing broke nav (it would open and close but no longer would responding to navigation clicks or any model-view updates) 

no errors were thrown when de-serializing. I'll post an example in slack

dkrasner:
copy and pasted elements don't appear in the nav

here is example in an area 
image

but also true of any place you can paste (like a card)
image

dkrasner:
also [the dev tools message func](https://github.com/UnitedLexCorp/SimpleTalk/blob/eric-nav-ref1/js/objects/System.js#L184) is throwing errors. Do you want to just disable it [here](https://github.com/UnitedLexCorp/SimpleTalk/blob/eric-nav-ref1/js/objects/System.js#L228)

i did this on my branch

we can come back to message related devtools, but as discussed it might be time to move all this debugging to ST itself

dkrasner:
@darth-cheney 

i think this is 


i am still/now? seeing this issue: opening script editors on any view doesn't appear in the nav card view, but does in the nav stack view. This is also true for the `EditorView` custom component but that is b/c it's not an ST part (when we fix that up one way or another this will go away) 
image

ApproximateIdentity:
One small thing I noticed is that if I save a page with the navigator open, it doesn't stay open when I open that page. So I guess the state of the navigator pane isn't included in the serialization. I assume we don't care about this at all, but if it's a trivial thing to add it might be good since it makes the idempotency a little stronger. But super small and not too important.

dkrasner:
> One small thing I noticed is that if I save a page with the navigator open, it doesn't stay open when I open that page. So I guess the state of the navigator pane isn't included in the serialization. I assume we don't care about this at all, but if it's a trivial thing to add it might be good since it makes the idempotency a little stronger. But super small and not too important.

@ApproximateIdentity yeah this part of the design. Otherwise it won't load on de-serialization

PR #184, opened 2021-04-13
Cleanup

dkrasner:
### Main points ###

This PR is a general clean:

removing all drag and drop related code
removing all 'old' copy and paste related code
removing unnecessary handlers from system
refactoring system handlers (like openScriptEditor, openGrammar, and openDebugger) to be simpler and more ST like
removing weird injected js script in the serialized examples

Note: i made script editor an area with 4 suparts (title field, script field and the save button). This seems more natural, and allows for more layout options than a window but we might or might not want. If we keep this, we prob want to fix the resizing of the entire area which doesn't seem to work in the script editor (even though you can resize the underlying fields and everything is set to width 100%)


darth-cheney:
Looks good. Tests are passing. I don't see anything unusual when using. Also -2k lines of code, excellent.
  
The only question I have is the script editor. Right now it appears to just be an Area, which is fine in theory. But in use it's a little clumsy: you have to select the right halo to move it and you need to also use the halo to close it, is that right?

dkrasner:
yeah that's right but i played around with it and i don't like it at all... 

so i reverted to the original window setup, but cleaned that up. I also put the editor on the current card - we had it on the current stack which is why the script editors kept hanging around when you navigated to the next card. For this i needed to allow cards to accept windows as subparts, which I think is right but for some reason we decided against it earlier.

Issue #187, opened 2021-04-14
Hand recognition doesn't quite save right


ApproximateIdentity:
The way the hand recognition interface works is that it creates the lenin hand when it boots up, moves it around when running, and then removes it again when it turns off. But the hand itself is just an st-image. If you save the page when the interface is running it entirely ignore the fact that the interface is running (which is probably good actually), but as a result it considers the st image as an object that needs saving. Once I reload the saved page, it creates the hand image again as a regular image, but if I re-start the hand recognition it creates a new image. The old one is just a random ST object floating around.

In other words, my previous assumption was that if hand recognition turns off, it destroys the lenin hand image, but that is not true if I save it and reload it. So the serialize/de-serialize process sort of creates an object leak with respect to my assumption. I am not entirely sure what the best way is to fix this. It might be to create some sort of unique id for the hand recognition and check if it exists. Another possible solution is to have the serialization process not save the hand itself. Gotta think of this a little...

dkrasner:
This hand recognition on/off should be a part property. 

There are some options here:

`st-image`, or really any part that moves, could have part property called something like `recognizeHand: true/false` . This can be a dynamic property which when sets to `true` loads up the hand rec. related module; when set to `false` shuts it all down. 

and for example the `Hand Recognition` button in toolbox which now calls the system command `toggleHandDetection` would do something like:

```
on click
     add image "Lenin"
     tell image "Lenin" to set "src" to [uri]
     tell image "Lenin" to set "recognizeHand" to true  -- alternatively  we can add a private command called `recognizeHand` so `tell image "Lenin" to recognizeHand`
end click

On deserialization, as the parts/views are added and properties are set, the hand recognition would automatically turn on. This is really the only way we have to preserve state and keep with the current design for everything else in the system. 


 alternatively `recognizeHand` could be a property of card, stack or world doing more or less the same thing...

PR #190, opened 2021-04-15
Make Lenin hand a stack-level object

ApproximateIdentity:
This means that the hand follows around on different cards as expected. It also grabs onto objects on the current card and leaves them on the current card if cards are switched (which sounds confusing, but is what you would expect).

This does not make the hand a world level object.

### DIFF
--- a/js/objects/System.js
+++ b/js/objects/System.js
@@ -893,6 +893,11 @@ System._commandHandlers['openDebugger'] = function(senders, partId){
 };
 
 System._commandHandlers['saveHTML'] = function(senders){
+    // Stop hand recognition if it's running.
+    let handRecognitionOriginallyRunning = handInterface.handDetectionRunning;
+    if (handRecognitionOriginallyRunning) {
+        handInterface.stop();
+    }
     this.serialize();
 
     let stamp = Date.now().toString();
@@ -908,6 +913,10 @@ System._commandHandlers['saveHTML'] = function(senders){
     anchor.click();
     window.URL.revokeObjectURL(url);
     anchor.parentElement.removeChild(anchor);
+    // Start hand recognition if it was running.
+    if (handRecognitionOriginallyRunning) {
+        handInterface.start();
+    }
 };
 
 System._commandHandlers['tell'] = (senders, targetId, deferredMessage) => {
--- a/js/objects/System.js
+++ b/js/objects/System.js
@@ -803,22 +803,6 @@ System._commandHandlers['importWorld'] = function(sender, sourceUrl){
                         // Tell the world to update current, in case
                         // a new StackView was attached with current set.
                         document.querySelector('st-world').updateCurrentStack();
-                        // This moves all image children of world to the end preserving stack order.
-                        const world = document.querySelector('st-world');
-                        var imageChildren = [];
-                        for (var i = 0; i < world.children.length; ++i ) {
-                            const child = world.children[i];
-                            if (child.localName == 'st-image') {
-                              imageChildren.push(child)
-                            }
-                        }
-                        // Appending a child to the dom that already exists is apparently equivalent to a move. See here:
-                        //
-                        //      https://stackoverflow.com/questions/34685316/reorder-html-elements-in-dom/51704111#51704111
-                        for (var i = 0; i < imageChildren.length; ++i) {
-                            const child = imageChildren[i];
-                            world.appendChild(child);
-                        }
                     });
                 };
             });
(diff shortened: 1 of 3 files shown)


ApproximateIdentity:
Okay the last commit makes the hand world-level and it works in all the ways you would want _except_ that the hand is invisible on other stacks. But the hand is working on other stacks because it will correctly grab buttons on those stacks and move them around.

@dkrasner Maybe this is related to the object ordering thing you were talking about on the call?

dkrasner:
@ApproximateIdentity yep that's the reason

you can always see if that's the case in the dev tools inspector. And a nice feature of it is that you can reorder the DOM elements with a drag&drop right there

image


 i just dragged that (last-child) stack to be ahead of the hand image and all was fine

do you want me to add a fix on this branch so that stacks always come in order?

js/objects/System.js
```diff
@@ -846,6 +846,36 @@ System._commandHandlers['importWorld'] = function(sender, sourceUrl){
                         // Tell the world to update current, in case
                         // a new StackView was attached with current set.
                         document.querySelector('st-world').updateCurrentStack();
+                        // This reorders all children of world so that non-stacks are at the end.
```
dkrasner, line 849: this shouldn't need to happen on import

if we make sure stacks are always added in order, like cards, then this will be preserved on serialization (and hence de-serialization)

ApproximateIdentity:
I think that your version just plain doesn't work. I rebased this branch on master and it continues to work (though using my method that maybe we don't like). I also pushed a branch `thomas-lenin-hand-stack-daniel-version` here https://github.com/UnitedLexCorp/SimpleTalk/tree/thomas-lenin-hand-stack-daniel-version which is your commit also rebased on master. In other words, these two branches are the same except for the last commit.

Yours is not working for me. I follow these steps to reproduce:

1. Start with bootstrap.html
2. Turn on hand detection
3. Import world bootstrap.html again
4. Look at the new screen and it is not there

That works on my branch (i.e. the hand stays on top on all stacks), but not on yours. I ran those steps again and put a breakpoint on the code here where your last commit made the changes:

https://github.com/UnitedLexCorp/SimpleTalk/blob/thomas-lenin-hand-stack-daniel-version/js/objects/System.js#L465

So the function that you made your changes in is never being called by the process.

Maybe @darth-cheney can check and verify that he's seeing the same thing?

ApproximateIdentity:
@dkrasner I changed this to only care about `st-image` objects and nothing else. You cool with this?

dkrasner:
> @dkrasner I changed this to only care about `st-image` objects and nothing else. You cool with this?



> I think that your version just plain doesn't work. I rebased this branch on master and it continues to work (though using my method that maybe we don't like). I also pushed a branch `thomas-lenin-hand-stack-daniel-version` here https://github.com/UnitedLexCorp/SimpleTalk/tree/thomas-lenin-hand-stack-daniel-version which is your commit also rebased on master. In other words, these two branches are the same except for the last commit.
> 
> Yours is not working for me. I follow these steps to reproduce:
> 
>     1. Start with bootstrap.html
> 
>     2. Turn on hand detection
> 
>     3. Import world bootstrap.html again
> 
>     4. Look at the new screen and it is not there
> 
> 
> That works on my branch (i.e. the hand stays on top on all stacks), but not on yours. I ran those steps again and put a breakpoint on the code here where your last commit made the changes:
> 
> https://github.com/UnitedLexCorp/SimpleTalk/blob/thomas-lenin-hand-stack-daniel-version/js/objects/System.js#L465
> 
> So the function that you made your changes in is never being called by the process.
> 
> Maybe @darth-cheney can check and verify that he's seeing the same thing?

@ApproximateIdentity  yes b/c you are testing the import of an _already existing_ world, as opposed to adding parts to world

So this works exactly as intended. On the  `thomas-lenin-hand-stack-daniel-version` branch I just ran 

image


note the added stack comes before the image, even though the opposite order commands was run. 


Again this is nothing to do with importing worlds. Those should just come in as is. 


Perhaps the best thing to do is to stop the image recognition if a world is imported and then start it again. This is simple and allow us to see how things play out, what kinds of parts we'll want on world or anywhere else etc

ApproximateIdentity:
@dkrasner I changed it so that importing world stops and restarts hand recognition if it's running. i also added a second commit to do essentially the same thing when saving a world which should satisfy this issue: https://github.com/UnitedLexCorp/SimpleTalk/issues/187

thoughts about this method?

dkrasner:
@ApproximateIdentity i think this is fine for now. In the future all this card and stack priority will need to be solved with some sort of `slot`'ing   otherwise we'll end up chasing issues like this all over the place (as opposed insisting in one place where cards and stacks go - i'll work on this in the next little while)

ApproximateIdentity:
okay cool. yeah i think the only reason this is sort of acceptable for the moment is because the hand is now the only example of this. but the second we want to do something else similar, it definitely needs a less one-off solution.

Issue #200, opened 2021-04-20
Drawing size is not preserved


dkrasner:
the size of the drawing part-view is not preserved on copy&paste or serialization

this is b/c drawing re-size [does not use](https://github.com/UnitedLexCorp/SimpleTalk/blob/master/js/objects/views/drawing/DrawingView.js#L231) part properties

PR #205, opened 2021-04-21
Various Fixes

### LINKED ISSUE #198: Image width/height are not preserved
Here is an example:

add image to card and set its width and height to "fill" (i.e. 100%)

save, then load

the image will have the original (import image) dimensions

dkrasner:
### Main Points ###

Fixing some bugs, tying up loose ends:

* MNU error handler was not working when combining `do to` with a `tell` command. Now gets the proper context from the execution stack. 
* image width and height were not being routed properly through props, preventing serialization and copying functionality. (@darth-cheney I commented out [this line](https://github.com/UnitedLexCorp/SimpleTalk/compare/daniel-various-fixes?expand=1#diff-f23fdb301be3994e39028ed283c38bcbb22e821fc23b7e90c75e716fed0d8b14R119) since now this is handled by the combination of inline and class defined css, and everything seems to work. With that line there, the image is resized on load based on what the view thinks is appropriate, screwing up ST width/height props -> css)
* drawing width and height were not being routed properly through props... 
* cleaning up various unused methods 

(adding language.html <- starting point for our language stack)

this resolve #198 and #200 

Play around making images and drawings, saving, loading and copying.


darth-cheney:
There is a strange issue with preserving the aspect ratio when resizing binary images, as demonstrated by these two screenshots:
Resize1
Resize2

PR #217, opened 2021-04-30
Delete command cleanup

### LINKED ISSUE #213: "delete" command should use object specifiers
### Main Points ###

our current grammar/semantics implementation of the "delete" command does no use object specifiers, restricting the grammar and making some of the handling a bit more complex than necessary. This is due to the fact these commands were written before object specifiers were fully developed. 

The "deleteModel" command handler should also be removed from part.

dkrasner:
### Main Points ###

This is another small PR, similar to #212, which updates our `delete` command to use object specifiers. 

Now you can write
```
delete this button
delete first button ...
delete button id .
etc
```

This also removes the delete command handler from Part, which dealt with some context related de-referencing before proper object specifier semantics was written. 

Tests have been updated. 

Closes #213


darth-cheney:
I've tested this a bit and ran all the automated tests, which are passing. I say merge it in.

darth-cheney:
@dkrasner One thing we need to add here is an event to notify the Navigator when something has been removed, ie the equivalent of this (from adding a view): https://github.com/UnitedLexCorp/SimpleTalk/blob/daniel-delete-cleanup/js/objects/System.js#L485
  
I can push the required changes later today

dkrasner:
> @dkrasner One thing we need to add here is an event to notify the Navigator when something has been removed, ie the equivalent of this (from adding a view): https://github.com/UnitedLexCorp/SimpleTalk/blob/daniel-delete-cleanup/js/objects/System.js#L485
> 
> I can push the required changes later today
@darth-cheney yeah do you want to do that and then merge it in?
> 
> Additionally, we should decide what is supposed to happen if you make a button that has `delete this card`. Should we navigate to the previous or next card? What if there is only one card in the current stack -- do we do nothing, or delete the card and create a new blank one automatically?

I think that you should not be able to delete the only card on a stack. This will resolve some of the ambiguities. Otherwise navigating to the next card is good.

darth-cheney:
@dkrasner Ok I pushed the event/changes so that Nav now responds to Card and Stack removals. Test it out and you'll see what happens. See commit message for more info.

dkrasner:
@darth-cheney I played around with ^^ and the nav stuff is working fine. I am not a fan of having a mix of web events and ST messages and eventually think we'll need to rewrite nav into something more ST native. Otherwise we'll always end up catching/hacking nav to work with whatever the latest ST changes are. 

There is a bigger problem though with delete. The `current` property is an index which means that our state depends on the number of current cards or stacks at the moment of the change. This works fine if your are only adding but when removing it causes race conditions. 

For example, 
if you remove the current card and you want to navigate to the next one the following occurs:

going to the next card depends on knowing what the current one is, so you can't delete it first
going to the next card will move the index of `current` + 1 modulo (total cards), which won't be the same after you delete the current card, i.e. the modulus will change, so you can't delete it after

You can definitely hack some logic into System.deleteModel which will account for this, dealing with modulus changes, going to the right card/stack by id if necessary. But that's no good. 

You can also insist that removing the current card or stack is not allowed, which is what I did in the last commit there. 

But regardless, I think that having a property that looks like its designed for one operation and not it's inverse (new vs delete) should be avoided. We should make `current` an id not an index. 

What do you think?

dkrasner:
@darth-cheney you good to merge this?

PR #218, opened 2021-05-01
refactoring audio

dkrasner:
### Main Points ###

The one commit here, came out of my "warming up" for the remote resource work. It seemed reasonable to see what `st-audio` will need at a minimum to be viable and how that would apply to a more general `st-resource` part/view. 

**But** this has made `st-audio` much nicer in my view. It acts a lot more like a cross between `st-image` and `st-button` - it has its own default icon and displays a name, allows for style and text-style props etc. And it gives some feedback (border turns green when the resource is loaded). 

So since this is good for demos (I made a pretty cool player with multiple `st-audio` parts "tracks" and used the halo target on a button to flip between them), I figured it might make sense to merge this now as I continue to work on the more general thing.


ApproximateIdentity:
Maybe this PR is a good excuse to update `examples/audio-testing.html` so that it works again?

dkrasner:
> Maybe this PR is a good excuse to update `examples/audio-testing.html` so that it works again?

yeah i'll make a new one

dkrasner:
@ApproximateIdentity [here](https://github.com/UnitedLexCorp/SimpleTalk/pull/218/commits/1fcb1c2911babce1a55281eb82be841a8822f844) it is. Give a go..

there was a lifecycle bug in the resource loading, which I didn't catch until I made a new example snapshot.

ApproximateIdentity:
Works well for me! Seems good.

darth-cheney:
This looks good to me and is working great. Only thing I would add is to make sure we set `cursor: pointer` for the new AudioView elements in the styling somewhere, so that it's clear these are clickable.

PR #222, opened 2021-05-04
Stack order

dkrasner:
### Main Points ###

Stacks, like cards, should be added in order. 

Technically we could use slots for this, but this PR would be a temp fix.


darth-cheney:
Looking good from my end. I think  it's good to merge.

